PR #28, opened 2014-07-12
Allow unintrusive customization of CUB's top-level namespace

jaredhoberock:
The current way `util_namespace.cuh` works requires the user to edit it to customize `CUB_NS_PREFIX` & `CUB_NS_POSTFIX`. It's useful to allow the user to simply `#define` it in his source code before including a CUB header. This patch permits that use case.


allisonvacanti:
Closing PRs that are several years old with unintelligible diffs. Looks like maybe the repo's history was rewritten at some point.

PR #125, opened 2018-02-01
Added missing 'template' keyword.

ilya-biryukov:
To unbreak compilation with clang.


ilya-biryukov:
For context: we need this change to unbreak compilation of tensorflow with cuda-clang configuration

allisonvacanti:
This was fixed by 53daec7.

PR #155, opened 2018-12-08
fix 'invalid arguments' warp sync error on Volta

jglaser:
Fixes issue #154


GLJeff:
Is something like DeviceRadixSort even safe to run right now without your fix?

allisonvacanti:
Looks like this was closed accidentally during the repo move. Reopening.

allisonvacanti:
I'd like to understand a bit more about the usecase where this would be an issue before addressing this. Backlogging until a repro that demonstrates the issue is available (see [this comment](https://github.com/NVIDIA/cub/issues/154#issuecomment-867196197) for details).

PR #164, opened 2019-05-16
Use .sync variants of shfl/vote for CUDA 9+.

Artem-B:
Non-sync variants have been deprecated and are now completely gone
in CUDA-10.1 when we compile for sm_7x.

### DIFF
--- a/cub/util_ptx.cuh
+++ b/cub/util_ptx.cuh
@@ -285,6 +285,8 @@ __device__  __forceinline__ int WARP_ANY(int predicate, unsigned int member_mask
 {
 #ifdef CUB_USE_COOPERATIVE_GROUPS
     return __any_sync(member_mask, predicate);
+#elif CUDA_VERSION >= 9000
+    return __any_sync(0xffffffff, predicate);
 #else
     return ::__any(predicate);
 #endif
@@ -298,6 +300,8 @@ __device__  __forceinline__ int WARP_ALL(int predicate, unsigned int member_mask
 {
 #ifdef CUB_USE_COOPERATIVE_GROUPS
     return __all_sync(member_mask, predicate);
+#elif CUDA_VERSION >= 9000
+    return __all_sync(0xffffffff, predicate);
 #else
     return ::__all(predicate);
 #endif
@@ -311,6 +315,8 @@ __device__  __forceinline__ int WARP_BALLOT(int predicate, unsigned int member_m
 {
 #ifdef CUB_USE_COOPERATIVE_GROUPS
     return __ballot_sync(member_mask, predicate);
+#elif CUDA_VERSION >= 9000
+    return __ballot_sync(0xffffffff, predicate);
 #else
     return __ballot(predicate);
 #endif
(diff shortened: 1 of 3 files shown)


Artem-B:
Ping. Is there anybody out there?

dumerrill:
Looking after it :)

dumerrill:
I'm not sure what the point of these edits is: CUB already uses the *sync versions exclusively when compiled with CUDA tookits >= 9x.  This is enabled by the `CUB_USE_COOPERATIVE_GROUPS` macro, which is defined https://github.com/NVlabs/cub/blob/master/cub/util_arch.cuh#L48:

```
#if (__CUDACC_VER_MAJOR__ >= 9) && !defined(CUB_USE_COOPERATIVE_GROUPS)
    #define CUB_USE_COOPERATIVE_GROUPS
#endif
```

Further, blindly using `0xffffffff` would be inappropriate, as that could mean a hang when `member_mask` is a subwarp.

Artem-B:
Oh! I think I see what the problem is. We're compiling all CUDA sources with clang and __CUDACC_VER_* is only defined by NVCC. Enabling CUB_USE_COOPERATIVE_GROUPS should indeed make this patch unnecessary.

In general, is there particular reason for using nvcc-specific __CUDACC_VER* macros for checking *CUDA SDK* version? Shouldn't CUDA_VERSION be used instead? Granted, nvcc used to be the only CUDA compiler, and those macros would produce equivalent results, but this is no longer the case. 

> Further, blindly using 0xffffffff would be inappropriate, as that could mean a hang when member_mask is a subwarp.

AFAICT shfl will not hang. In diverged context it will produce undefined results for inactive threads. shfl.sync has the same behavior -- with all-ones mask, shfl.sync will also result in undefined data for inactive threads. For all practical purposes both old and new code assume that execution is non-diverged. There's no practical way to guarantee that in practice -- divergent branch may have happened way up the call chain where the code had no idea that one of the functions at the bottom may require non-diverged context. The bottom line is that the new assembly should be functionally equivalent to the one it replaces.

E.g. PTX ISA says this for for `shfl`:
> Note that results are undefined in divergent control flow within a warp, if an active thread sources a register from an inactive thread.

for `shfl.sync`:
> Note that results are undefined if a thread sources a register from an inactive thread or a thread that is not in membermask.

`vote` will produce 0 for inactive threads, so there's no hang either.

I may be missing something, though.

dumerrill:
Hi Artem,

I'll look into replacing the *CC macro with CUDA_VERSION.

For SM7+, the *sync() methods can hang if your mask requires certain
threads to show up, yet the application's control flow prevents them from
doing so.  (The compiler literally inserts a spin loop to wait for them all
to show up.)  For example, suppose some control flow in which lane0 skips a
shuffle in which lane31 shares a value with other threads:

if (threadIdx.x > 0)
   y = __shfl_sync(0xffffffff, x, 31);   // inappropriate mask (hang)
__warp_sync(0xffffffff);                 // all threads reconverge here


will deadlock because the other threads will wait indefinitely for thread0
to either (a) show up at the shuffle (per the mask), or (b) exit the
program, which it can't do until the warpsync is satisfied, which requires
the other threads waiting on the shuffle.

Secondly, the particular text you're referencing from the PTX ISA has to do
with what happens when you *have a proper mask*, but try to *source data
from an inactive lane*, e.g.:

if (threadIdx.x > 0)
   y = __shfl_sync(0xfffffffe, x, 0);   // appropriate mask (no hang), but
undefined value of y

Hope that helps,

Duane

________________________________________
Duane G Merrill III, PhD
Principal Research Scientist
NVIDIA Corporation
dumerrill@nvidia.com

Artem-B:
OK. Thank you for clarification.

PR #173, opened 2019-10-11
CUDA attribute caching

brycelelbach:
This change adds a multi-device aware caching facility for `PtxVersion` and `SmVersion`. CUDA attribute queries, such as `cudaFunctionGetAttributes` and (to a lesser degree) `cudaDeviceGetAttribute` call into the CUDA driver and may involve the acquisition of driver locks, which can cause slowness, especially under contention. Some users (RAPIDS, FB PyTorch) have encountered performance issues because of this.

### DIFF
--- a/cub/util_device.cuh
+++ b/cub/util_device.cuh
@@ -55,7 +55,7 @@ namespace cub {
 
 
 /**
- * Alias temporaries to externally-allocated device storage (or simply return the amount of storage needed).
+ * \brief Alias temporaries to externally-allocated device storage (or simply return the amount of storage needed).
  */
 template <int ALLOCATIONS>
 __host__ __device__ __forceinline__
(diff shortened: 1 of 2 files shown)


cub/util_device.cuh
```diff
@@ -158,28 +222,81 @@ CUB_RUNTIME_FUNCTION __forceinline__ cudaError_t PtxVersion(int &ptx_version)
 #endif
 }
 
+/**
+ * \brief Retrieves the PTX version that will be used on \p device (major * 100 + minor * 10).
+ */
+CUB_RUNTIME_FUNCTION __forceinline__ cudaError_t PtxVersionUncached(int &ptx_version, int device)
+{
+    SwitchDevice sd(device);
```
griwes: This makes twice as many `cudaSetDevice` calls as needed, doesn't it? Because effectively it jumps back and forth, instead of going to 0, 1, 2, 3, ..., and only then back to the original one?

I'm not sure if this is significant, just pointing that out.

allisonvacanti:
Is this still needed?

brycelelbach:
No this has been merged.

PR #185, opened 2020-06-01
Add assignment operator to the TestBar test util class.

hwinkler:
n.b. this change to test code introduces a lot of failures in the CUB unit tests.
I'm not submitting those fixes -- I've done a few but not all of them. 

This change will identify all places in CUB code that attempts an assignment to uninitialized memory.
See https://github.com/NVlabs/cub/issues/184

Change:
I added a an assignment operator to the class, and I added a field
named `magic` that I set to certain value during
construction. Only an object that has been constructed is likely to
have that magic value properly set.

An assignment precondition is that the left-hand side is valid.
The assignment operators I added assert that the `magic` member of
the lhs has the correct value.

### DIFF
--- a/test/test_util.h
+++ b/test/test_util.h
@@ -1140,6 +1140,7 @@ struct NumericTraits<TestFoo>
 // Complex data type TestBar (with optimizations for fence-free warp-synchrony)
 //---------------------------------------------------------------------
 
+#define MAGIC 895245
 /**
  * TestBar complex data type
  */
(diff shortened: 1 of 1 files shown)


allisonvacanti:
Converting to draft to prevent merging since this PR's purpose is to detect errors, not fix them.

test/test_util.h
```diff
@@ -1140,29 +1140,41 @@ struct NumericTraits<TestFoo>
...
 
     // Constructor
-    __host__ __device__ __forceinline__ TestBar(long long x, int y) : x(x), y(y)
+    __host__ __device__ __forceinline__ TestBar(long long x, int y) : x(x), y(y), magic(MAGIC)
     {}
 
+    // Assignment operator
+    __host__ __device__ __forceinline__ TestBar& operator =(const TestBar& that)
+    {
+        assert (magic == MAGIC);
```
brycelelbach, line 1168: This should probably be something other than `assert` - in a release build, asserts may be turned off, and we want to catch cases where miscompiles in -O3 lead to missed assignments as well. Perhaps it should use the CUB error reporting machinery.

PR #204, opened 2020-09-24
Faster Least Significant Digit Radix Sort Implementation

canonizer:
- radix sort with decoupled look-back, 8 bits per pass and other optimizations
- pull request to the previous CUB repository: https://github.com/brycelelbach/cub_historical_2019_2020/pull/26


allisonvacanti:
Looks good on nvc++, now retesting DVS CL 29114896.

allisonvacanti:
There are still some issues with our internal testing system that are preventing me from ok'ing this just yet. Hopefully this will be resolved soon.

allisonvacanti:
DVS-AUS only CL: 29227867

allisonvacanti:
Rebased and squashed in last push.

Another DVS-AUS CL: 29243517

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1113,6 +1363,141 @@ struct DispatchRadixSort :
...
+    template <typename ActivePolicyT>
+    CUB_RUNTIME_FUNCTION __forceinline__
+    cudaError_t InvokeOnesweep()
+    {
+        typedef typename DispatchRadixSort::MaxPolicy MaxPolicyT;
+#if defined(__NVCOMPILER_CUDA__)
+        typedef OffsetT AtomicOffsetT;
+#else
+        typedef cuda::atomic<OffsetT, cuda::thread_scope_device> AtomicOffsetT;
+#endif
```
allisonvacanti, line 1375: There's still an issue using libcu++ atomics in this patch:

```
#if defined(__CUDA_ARCH__) && ((!defined(_MSC_VER) && __CUDA_ARCH__ < 600) || (defined(_MSC_VER) && __CUDA_ARCH__ < 700))
#  error "CUDA atomics are only supported for sm_60 and up on *nix and sm_70 and up on Windows."
#endif
```

Since this is host code, we need this to work across all SM versions.

It looks like we'll need to move the `AtomicOffsetT` into the per-SM policies, accounting for the different minimum version on windows. However, that check is at file-scope in `cuda\std\detail\__atomic`, so just including the atomic header while targeting older SMs is problematic.

I'll check with the libcu++ folks about this, maybe they know of a workaround.

cub/agent/agent_radix_sort_onesweep.cuh
```diff
@@ -219,11 +211,8 @@ struct AgentRadixSortOnesweep
                 // write the local sum into the bin
                 AtomicOffsetT& loc = d_lookback[block_idx * RADIX_DIGITS + bin];
                 OffsetT value = bins[u] | LOOKBACK_PARTIAL_MASK;
-#if defined(__NVCOMPILER_CUDA__)
-                (volatile OffsetT&)loc = value;
-#else
-                loc.store(value, cuda::memory_order_relaxed);
-#endif
+                //(volatile OffsetT&)loc = value;
+                ThreadStore<STORE_CG>(&loc, value);
```
allisonvacanti: Why not `STORE_VOLATILE` to match the previous behavior? Was volatile too strong here?

allisonvacanti:
DVS CL: 29264925.

allisonvacanti:
Rebased and squash, resubmitted DVS CL 29278629.

allisonvacanti (approved):

jlebar:
🎉   I'm excited to try this out.

For my info, under what circumstances should I expect to see a speedup with this patch?

elstehle:
Indeed, great work and congratulations on landing this! 👍 
@jlebar: @canonizer has presented this at GTC this year, here's a preview on the performance gains: 
https://developer.download.nvidia.com/video/gputechconf/gtc/2020/presentations/s21572-a-faster-radix-sort-implementation.pdf#page=20

Regarding "under what circumstances". From what I see in the tuning policies in [cub/device/dispatch/dispatch_radix_sort.cuh](https://github.com/NVIDIA/cub/pull/204/files#diff-25e00e1f1419432d6e0c5a8ad5968946931ca94e4aee13031163bf14f2b4a70f), it's for all GPU architectures of Pascal onwards, once the keys are at least 32 bits wide.

I'll leave it to the author to provide more details... 😃

canonizer:
Thanks @elstehle for summing it up. Yes, the new approach is now used for 32-bit and 64-bit keys for Pascal and above.

The main reason it is not enabled for earlier architectures is that I haven't done any performance experiments with them. If anyone does it and gets a speedup, they're welcome to send a pull request.

jlebar:
Thanks!

I ran some benchmarks on my RTX 2080 and didn't see a significant change for the <int, int> sort I'm doing.  It's a relatively small sort though, only 300k elements and not even sorting on all 32 bits.

jlebar:
Actually, I take it back -- PEBKAC.  I get a nice speedup from this change.  It's only zero speedup if I don't apply the patch.  :)

PR #207, opened 2020-09-30
Fixed CubDebug usage in CachingDeviceAllocator::DeviceAllocate

ahehn-nv:
Fixed a call to `CubDebug` in `CachingDeviceAllocator::DeviceAllocate`, where `CubDebug` was called on the return value of the `==` operator instead of the CUDA runtime API call.


allisonvacanti:
LGTM. Testing under DVS CL 29162721.

allisonvacanti (approved):

PR #213, opened 2020-10-14
Remove policies for arch < sm35.

allisonvacanti:
Goal is to clean up code and reduce build time:

Before:

| Target | Build Time (Wall) | Build Time (CPU seconds) |
|-|-|-|
| thrust.cpp.cuda.cpp17.all | 18:36.33 | 12414 |
| cub.cpp17.all | 22:58.27 | 4872 |

After:

| Target | Build Time (Wall) | Build Time (CPU seconds) |
|-|-|-|
| thrust.cpp.cuda.cpp17.all | 18:18.72 | 12174 |
| cub.cpp17.all | 13:53.14 | 3943 |

Thrust isn't significantly affected, as it spends most of its time building non-CUB algorithms (the set operation and minmax tests in particular). We can probably bring that down by removing old SM policies in a follow up patch, though Thrust currently only appears to go down to sm30 so there are fewer policies to take out.

CUB's compile time is reduced by nearly 40%. There's still a *lot* of room for improvement by looking at the CPU vs walltime (build used 12 cores). This is because there are one or two tests that take a very long time to compile. We can improve this by splitting up those tests so they can be parallelized.


allisonvacanti:
DVS CL 29199322

griwes (approved):

allisonvacanti:
Starting DVS on CL 29260540.

🤞

allisonvacanti:
No timeouts :tada:

PR #217, opened 2020-10-16
Update install rules to install cmake into lib directory

kkraus14:
See https://github.com/nvidia/thrust/issues/1322

Updates the cmake install rules to install in `lib/cub/cmake` instead of `include/cub/cmake` as the include folder is not searched in a normal `find_package(cub)` call.

### DIFF
--- a/cub/cmake/cub-config-version.cmake
+++ b/cub/cmake/cub-config-version.cmake
@@ -1,5 +1,12 @@
 # Parse version information from version.cuh:
-file(READ "${CMAKE_CURRENT_LIST_DIR}/../../../include/cub/version.cuh" CUB_VERSION_HEADER)
+unset(_CUB_VERSION_INCLUDE_DIR CACHE) # Clear old result to force search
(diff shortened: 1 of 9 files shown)


cub/cmake/cub-config-version.cmake
```diff
@@ -1,5 +1,5 @@
 # Parse version information from version.cuh:
-file(READ "${CMAKE_CURRENT_LIST_DIR}/../version.cuh" CUB_VERSION_HEADER)
```
allisonvacanti, line 2: File extension needs to stay `.cuh`

allisonvacanti (commented):
Noticed an accidental change to a filename, but I'll take care of that during local testing.

cub/cmake/cub-config-version.cmake
kkraus14: Woops, maybe that's why things broke for me when changing this 😅

allisonvacanti (approved):

PR #218, opened 2020-10-17
Stable sorting order for -0.0 and +0.0 for float and double.

canonizer:
This modifies digit extraction in radix sorting to sort -0.0 and +0.0 in stable order.

### DIFF
--- a/cub/block/block_radix_sort.cuh
+++ b/cub/block/block_radix_sort.cuh
@@ -36,6 +36,7 @@
 
 #include "block_exchange.cuh"
 #include "block_radix_rank.cuh"
+#include "radix_rank_sort_operations.cuh"
 #include "../config.cuh"
 #include "../util_ptx.cuh"
 #include "../util_type.cuh"
@@ -76,7 +77,9 @@ namespace cub {
  *   bit-sequences of \p RADIX_BITS as radix digit places.  Although the direct radix sorting
  *   method can only be applied to unsigned integral types, BlockRadixSort
  *   is able to sort signed and floating-point types via simple bit-wise transformations
- *   that ensure lexicographic key ordering.
+ *   that ensure lexicographic key ordering. For floating-point types -0.0 and +0.0 are
+ *   considered equal and appear in the result in the same order as they appear in
+ *   the input.
  * - \rowmajor
  *
  * \par Performance Considerations
@@ -175,6 +178,9 @@ private:
             PTX_ARCH>
         DescendingBlockRadixRank;
 
+    /// Digit extractor type
+    typedef BFEDigitExtractor<KeyT> DigitExtractorT;
+
     /// BlockExchange utility type for keys
     typedef BlockExchange<KeyT, BLOCK_DIM_X, ITEMS_PER_THREAD, false, BLOCK_DIM_Y, BLOCK_DIM_Z, PTX_ARCH> BlockExchangeKeys;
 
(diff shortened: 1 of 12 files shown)


allisonvacanti:
@canonizer Have you looked into at the performance impact of this? If not, I can use this as a case study for my new benchmarking code, but that will probably mean this won't make the 1.11.0 release.

allisonvacanti:
This implementation looks good to me, and the performance difference is negligible (<1% difference on GV100 from @canonizer's benchmarks).

Though now I'm curious whether this is actually expected behavior for users, or if it's better to just let them pre-condition their data when they want these to be treated equivalently. I'll do some polls and confirm whether or not this is something folks expect.

allisonvacanti:
Looks like the stl already does standard sorts this way, so that's a nice, definitive answer :)

@canonizer Can you update the `DeviceRadixSort` documentation to specify this behavior, and add a regression test to `test/test_device_radix_sort.cu`? I'll take a look at the other backends in Thrust to see if we can provide this guarantee there, too.

allisonvacanti:
@canonizer Any updates on this? I'd need to land this by Friday to make the 1.11.0 release, otherwise we can slip to 1.12.0.

cub/block/block_radix_sort.cuh
```diff
@@ -76,7 +77,9 @@ namespace cub {
  *   bit-sequences of \p RADIX_BITS as radix digit places.  Although the direct radix sorting
  *   method can only be applied to unsigned integral types, BlockRadixSort
  *   is able to sort signed and floating-point types via simple bit-wise transformations
- *   that ensure lexicographic key ordering.
+ *   that ensure lexicographic key ordering. For floating-poitn types -0.0 and +0.0 are
```
allisonvacanti, line 80: typo, s/poitn/point/

cub/block/block_radix_sort.cuh
```diff
@@ -76,7 +77,9 @@ namespace cub {
...
+ *   considered equal and appear in the result in the same order as the appear in
```
allisonvacanti, line 81: s/as the/as they/

cub/block/block_radix_sort.cuh
```diff
@@ -141,6 +144,9 @@ private:
         // The thread block size in threads
         BLOCK_THREADS               = BLOCK_DIM_X * BLOCK_DIM_Y * BLOCK_DIM_Z,
 
+        // Whether the key is floating-point
+        FLOAT_KEY                   = Traits<KeyT>::CATEGORY == FLOATING_POINT,
```
allisonvacanti, line 148: This doesn't seem to be used.

cub/device/device_radix_sort.cuh
```diff
@@ -66,7 +66,9 @@ namespace cub {
  * half-precision floating-point type.  Although the direct radix sorting
  * method can only be applied to unsigned integral types, DeviceRadixSort
  * is able to sort signed and floating-point types via simple bit-wise transformations
- * that ensure lexicographic key ordering.
+ * that ensure lexicographic key ordering. For floating-poitn types -0.0 and +0.0 are
```
allisonvacanti, line 69: Same as above

cub/device/device_segmented_radix_sort.cuh
```diff
@@ -66,7 +66,9 @@ namespace cub {
  * half-precision floating-point type.  Although the direct radix sorting
  * method can only be applied to unsigned integral types, DeviceSegmentedRadixSort
  * is able to sort signed and floating-point types via simple bit-wise transformations
- * that ensure lexicographic key ordering.
+ * that ensure lexicographic key ordering. For floating-poitn types -0.0 and +0.0 are
```
allisonvacanti, line 69: Same as above

allisonvacanti (dismissed):

allisonvacanti:
This looks good to me, I'll start testing soon. Thanks Andy!

allisonvacanti:
Heads up -- I rebased and squashed this to make this work with our scripts. We don't handle merging `main` into topic branches very well.

allisonvacanti:
DVS CL: 29407401

allisonvacanti (approved):

PR #225, opened 2020-10-26
Fix to support wrapping CUB in a custom outer namespace with macros

### LINKED ISSUE #222: Compile fails with anonymous namespace CUB_NS_PREFIX
reproduce:
``` cuda
#define CUB_NS_PREFIX namespace {
#define CUB_NS_POSTFIX }
#include <cub/cub.cuh>
int main()
{
        int res = (int)cub::Equals<int, int>::VALUE;
        return res;
}
```

```
/workspace/abc/a.cu(6): error: "cub" is ambiguous

/workspace/abc/a.cu(6): error: expected an identifier

/workspace/abc/a.cu(6): error: expected a ";"
```

Seems that [these lines](https://github.com/NVIDIA/cub/commit/629f01ec7b4f660d293899ab84680fb7819ece42#diff-b5b693806476b576b12bffce3bbe3cd6dca2326f0544c5cfb55083ff81736867R54-R57) introduced the errors.

elstehle:
Fixes #222 

Verified using the following example from issue #222 

```
#define CUB_NS_PREFIX namespace {
#define CUB_NS_POSTFIX }
#include <cub/cub.cuh>
int main()
{
        int res = (int)cub::Equals<int, int>::VALUE;
        return res;
}
```

### DIFF
--- a/cub/util_namespace.cuh
+++ b/cub/util_namespace.cuh
@@ -48,6 +48,7 @@
 #endif
 
 // Declare these namespaces here for the purpose of Doxygenating them
+CUB_NS_PREFIX
 
 /*! \namespace cub
  *  \brief \p cub is the top-level namespace which contains all CUB
@@ -57,3 +58,5 @@ namespace cub
 {
 
 }
+
+CUB_NS_POSTFIX


allisonvacanti:
DVS CL: 29265328

allisonvacanti (approved):

PR #230, opened 2020-10-31
Remove TEST_ARCH, which is no longer needed

brycelelbach:


test/CMakeLists.txt
```diff
@@ -1,7 +1,3 @@
-# TODO investigate whether this is really needed:
-math(EXPR CUB_TEST_ARCH "${CUB_MINIMAL_ENABLED_ARCH} * 10")
```
allisonvacanti, line 2: IIRC this is the only use of `CUB_MINIMAL_ENABLE_ARCH`, this should update `cmake/CubCudaConfig.cmake` as well to remove these lines:

```

# Create a variable containing the minimal target arch for tests
list(SORT enabled_archs)
list(GET enabled_archs 0 CUB_MINIMAL_ENABLED_ARCH)
```

and all of the other usages of `enabled_archs` in that file, too.

allisonvacanti (changes requested):
There's some more code that can be removed. Don't want to pass up that opportunity! :)

test/CMakeLists.txt
brycelelbach: Fixed.

allisonvacanti (approved):

PR #233, opened 2020-11-05
Fix cub/util_type.cuh to not use 16-bit float with NVC++

dkolsen-pgi:
A recent commit changed several occurances of `__CUDACC_VER_MAJOR__ >= 9` in preprocessor expressions to `(__CUDACC_VER_MAJOR__ >= 9 || CUDA_VERSION >= 9000)`.  NVC++ defines `CUDA_VERSION` but not `__CUDACC_VER_MAJOR__`, so the commit changed the condition from false to true for NVC++.  The code inside those blocks doesn't work with NVC++, causing compilation errors.  Fix the regression by changing the expressions to `(__CUDACC_VER_MAJOR__ >= 9 || CUDA_VERSION >= 9000) && !__NVCOMPILER_CUDA__`.

### DIFF
--- a/cub/util_type.cuh
+++ b/cub/util_type.cuh
@@ -37,7 +37,7 @@
 #include <limits>
 #include <cfloat>
 
-#if (__CUDACC_VER_MAJOR__ >= 9 || CUDA_VERSION >= 9000)
+#if (__CUDACC_VER_MAJOR__ >= 9 || CUDA_VERSION >= 9000) && !__NVCOMPILER_CUDA__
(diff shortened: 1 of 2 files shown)


dkolsen-pgi:
I have tested the change cub/util_type.cuh with NVC++.  I have not tested the change to test/test_device_radix_sort.cu, because I don't have an easy way to do that.

allisonvacanti (approved):

allisonvacanti:
LGTM, DVS CL 29289339 is passing.

PR #244, opened 2020-12-05
Rename CUB test variants to be less confusing and add a `CUB_TEST_VARIANT` CMake option to control them.

brycelelbach:
Historically, each CUB test has had up to three different test variants, controlled by macro defines.
 
* *"thorough" or no variant:* Comprehensive unit testing which typically covers all builtin test types, all scenarios, and has support for repeating the tests multiple times (via a `--repeat` flag).
* *"quicker"*: Minimal unit testing intend to compile and run in a shorter period of time and may only cover a subset of test types and scenarios.
* *"quick"*: Performance comparisons of CUB primitives with other libraries (usually Thrust, but also NPP in some places) and against theoretical hardware capabilities (memory bandwidth, etc). When CUB was initially released, Thrust had its own CUDA backend, which CUB aimed to compete with. The Thrust backend was replaced with a CUB-based backend because CUB was successful in that competition; as such, a number of these comparisons are less meaningful than they once were. However, the performance comparisons relative to theoretical hardware capabilities remain valuable.

The names "quicker" and "quick" are a bit confusing for the uninitiated; you might think that they are two different levels of thoroughness, however in reality one is for benchmarking and one is for minimal testing.

My guess is that "quick" (the benchmarks) were added first, and then later the "quicker" variants were added. At the time "quicker" was added, it's likely no one realized it would be a bit confusing.

This PR renames "quicker" to "minimal" and "quick" to "benchmark".

It also adds a `CUB_TEST_VARIANT` CMake option which can be used to control which test variants are built.
* `ALL` (default and current behavior): Build all available variants.
* `THOROUGH`: If a test has multiple variants, only build the thorough variant.
* `BENCHMARK`: If a test has multiple variants, only build the benchmark variant 
* Add a `CUB_TEST_VARIANT` option which controls which test variants are built.

Originally I was going to call this option `CUB_TEST_WORKLOAD`, but I didn't want to confuse the workload terminology with the terminology used by gpuCI and Thrust multi-config.


brycelelbach:
run tests

CMakeLists.txt
```diff
@@ -48,8 +48,13 @@ endif()
 
 option(CUB_ENABLE_HEADER_TESTING "Test that all public headers compile." ON)
 option(CUB_ENABLE_TESTING "Build CUB testing suite." ON)
+set(CUB_TEST_VARIANT "ALL" CACHE STRING "Specify which CUB test variants should be built (options: ALL, THOROUGH, BENCHMARK, MINIMAL).")
```
allisonvacanti, line 51: Can we make this a series of booleans

```
CUB_TEST_VARIANT_THOROUGH
CUB_TEST_VARIANT_MINIMAL
CUB_TEST_VARIANT_BENCHMARK
```

that default to `ON`? It would allow us to just turn off `THOROUGH` variants but leave still have some testing + benchmarks.

CMakeLists.txt
```diff
@@ -48,8 +48,13 @@ endif()
...
 option(CUB_ENABLE_EXAMPLES "Build CUB examples." ON)
 
+if (NOT CUB_TEST_VARIANT MATCHES "^(ALL|THOROUGH|BENCHMARK|MINIMAL)$")
```
allisonvacanti, line 54: This is moot if we switch to booleans, but FYI there's a better way to do this:

```
  set_property(CACHE CUB_TEST_VARIANT PROPERTY STRINGS
    ALL THOROUGH BENCHMARK MINIMAL
  )
```

This will populate dropdown menus, etc in CMake GUIs.

test/Makefile
```diff
@@ -43,10 +43,10 @@ include ../common.mk
 
 # Testing mode option (quick/thorough)
 ifeq ($(quickertest), 1)
-	NVCCFLAGS += -DQUICKER_TEST
+	NVCCFLAGS += -DMINIMAL_TEST
```
allisonvacanti, line 46: This doesn't match the `#ifdef`s, should be `CUB_TEST_MINIMAL`.

test/Makefile
```diff
@@ -43,10 +43,10 @@ include ../common.mk
...
 	TEST_SUFFIX = quicker
 else ifeq ($(quicktest), 1)
-	NVCCFLAGS += -DQUICK_TEST
+	NVCCFLAGS += -DBENCHMARK
```
allisonvacanti, line 49: Same, should be `CUB_TEST_BENCHMARK`

test/CMakeLists.txt
```diff
@@ -60,27 +60,27 @@ function(cub_add_test target_name_var test_name test_src cub_target)
   )
 endfunction()
 
-# Sets HAS_QUICK_VARIANT / HAS_QUICKER_VARIANT / NO_VARIANTS to True/False in
+# Sets HAS_BENCHMARK_VARIANT / HAS_MINIMAL_VARIANT / NO_VARIANTS to True/False in
 # the calling scope.
 # Used to detect variants of unit tests depending on whether a source file
-# contains the strings "QUICK_TEST" or "QUICKER_TEST".
+# contains the strings "BENCHMARK" or "MINIMAL_TEST".
```
allisonvacanti, line 66: These should check for the full strings `CUB_TEST_BENCHMARK`, `CUB_TEST_MINIMAL` to avoid false positives.

Checking for `MINIMAL_TEST` won't match the new `#ifdef`s (`CUB_TEST_MINIMAL`).

test/CMakeLists.txt
```diff
@@ -60,27 +60,27 @@ function(cub_add_test target_name_var test_name test_src cub_target)
...
 function(cub_check_for_test_variants src)
   file(READ "${src}" data)
 
-  string(FIND "${data}" "QUICK_TEST" quick_loc)
-  set(HAS_QUICK_VARIANT False PARENT_SCOPE)
-  if (NOT quick_loc EQUAL -1)
-    set(HAS_QUICK_VARIANT True PARENT_SCOPE)
+  string(FIND "${data}" "BENCHMARK" benchmark_loc)
```
allisonvacanti, line 70: `CUB_TEST_BENCHMARK`

test/CMakeLists.txt
```diff
@@ -60,27 +60,27 @@ function(cub_add_test target_name_var test_name test_src cub_target)
...
+  set(HAS_BENCHMARK_VARIANT False PARENT_SCOPE)
+  if (NOT benchmark_loc EQUAL -1)
+    set(HAS_BENCHMARK_VARIANT True PARENT_SCOPE)
   endif()
 
-  string(FIND "${data}" "QUICKER_TEST" quicker_loc)
-  set(HAS_QUICKER_VARIANT False PARENT_SCOPE)
-  if (NOT quicker_loc EQUAL -1)
-    set(HAS_QUICKER_VARIANT True PARENT_SCOPE)
+  string(FIND "${data}" "MINIMAL_TEST" minimal_loc)
```
allisonvacanti, line 76: `CUB_TEST_MINIMAL`

test/CMakeLists.txt
```diff
@@ -91,40 +91,50 @@ foreach (test_src IN LISTS test_srcs)
   get_filename_component(test_name "${test_src}" NAME_WE)
   string(REGEX REPLACE "^test_" "" test_name "${test_name}")
 
-  # Some tests change behavior based on whether the compiler defs QUICK_TEST
-  # and/or QUICKER_TEST are defined. Detect these and build variants for each
+  # Some tests change behavior based on whether the compiler defs BENCHMARK
```
allisonvacanti, line 94: Strings need updating.

test/CMakeLists.txt
```diff
@@ -91,40 +91,50 @@ foreach (test_src IN LISTS test_srcs)
...
+  # and/or MINIMAL_TEST are defined. Detect these and build variants for each
   # configuration:
   cub_check_for_test_variants("${test_src}")
 
   foreach(cub_target IN LISTS CUB_TARGETS)
     if (NO_VARIANTS)
       # Only one version of this test.
       cub_add_test(test_target ${test_name} "${test_src}" ${cub_target})
+    elseif (NOT CUB_TEST_VARIANT STREQUAL "ALL")
+      string(TOLOWER CUB_TEST_VARIANT lowercase_cub_test_variant)
```
allisonvacanti, line 104: This can be removed and rolled into the logic below if we switch to booleans options:
1. wrap the `thorough` logic in an `if (CUB_TEST_VARIANT_THOROUGH)`
2. update the `HAS_` checks to `if (CUB_TEST_VARIANT_XXX AND HAS_XXX_VARIANT)`

test/CMakeLists.txt
```diff
@@ -91,40 +91,50 @@ foreach (test_src IN LISTS test_srcs)
...
+      target_compile_definitions(${test_target_${lowercase_cub_test_variant}}
+                                 PRIVATE CUB_TEST_${CUB_TEST_VARIANT})
     else()
       # By default (no flags), the "thorough" version of the test is built:
       cub_add_test(test_target_thorough
         ${test_name}.thorough
         "${test_src}"
         ${cub_target}
       )
+      target_compile_definitions(${test_target_thorough} PRIVATE CUB_TEST_THOROUGH)
```
allisonvacanti, line 119: Nothing defines this, is this addition intentional?

allisonvacanti: If we add this here, it should be added to the Makefile, too.

test/test_device_radix_sort.cu
```diff
@@ -1220,14 +1220,14 @@ int main(int argc, char** argv)
...
-#ifdef QUICKER_TEST
+#ifdef CUB_TEST_MINIMAL
 
     enum {
         IS_DESCENDING   = false
     };
 
     // Compile/run basic CUB test
-    if (num_items < 0)      num_items       = 48000000;
+    if (num_items < 0)      num_items       = 24000000;
```
allisonvacanti, line 1230: This change looks unrelated. Was it intentional?

allisonvacanti (changes requested):
The explanation and overall change is great.

I'd like to tweak the implementation to simplify some of the logic and provide more flexibility over which tests get generated, see inline comments.

CMakeLists.txt
brycelelbach: Done.

CMakeLists.txt
brycelelbach: Done.

test/Makefile
brycelelbach: Can we just delete the makefiles at this point? Who uses them? We don't.

test/CMakeLists.txt
brycelelbach: Good catch.

test/CMakeLists.txt
brycelelbach: Fixed

test/CMakeLists.txt
brycelelbach: Fixed.

test/CMakeLists.txt
brycelelbach: Fixed.

test/CMakeLists.txt
brycelelbach: Fixed in a slightly different way.

test/test_device_radix_sort.cu
brycelelbach: This is intentional; that test just takes too long right now.

test/CMakeLists.txt
brycelelbach: Done. I just added it for completeness.

test/Makefile
allisonvacanti: Burn them 🔥🔥🔥

allisonvacanti (approved):
Looks good, approved.

Only major issue I have is that I'd still like to leave the suffixes on the test names for the single variant builds. It will make our lives easier when we need to debug CI problems and bug reports. What do you think?

test/CMakeLists.txt
```diff
@@ -60,71 +60,97 @@ function(cub_add_test target_name_var test_name test_src cub_target)
...
+  # Some tests change behavior based on whether the compiler defs BENCHMARK
+  # and/or MINIMAL_TEST are defined. Detect these and build variants for each
   # configuration:
   cub_check_for_test_variants("${test_src}")
 
   foreach(cub_target IN LISTS CUB_TARGETS)
     if (NO_VARIANTS)
       # Only one version of this test.
       cub_add_test(test_target ${test_name} "${test_src}" ${cub_target})
+    elseif (single_test_variant_option)
```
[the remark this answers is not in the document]
brycelelbach: That's probably fine, I'll yank it out.

test/CMakeLists.txt
brycelelbach: They are now.

CMakeLists.txt
```diff
@@ -48,6 +48,9 @@ endif()
 
 option(CUB_ENABLE_HEADER_TESTING "Test that all public headers compile." ON)
 option(CUB_ENABLE_TESTING "Build CUB testing suite." ON)
+option(CUB_TEST_VARIANT_THOROUGH "Build CUB thorough test variants." ON)
```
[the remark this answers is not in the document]
brycelelbach: Done.

Issue #228, opened 2020-10-29
CUB uses non-standard C++ extensions that break strict compilers


maddyscientist:
# Summary

Some Nvidia teams regularly use CUB with NVRTC, a compiler with a very strict interpretation of C++. CUB currently uses some non-standard language extensions, such as anonymous structs, which break NVRTC compilation. We should build our test harness with compiler flags that audit non-standard extension usages, such as `-Wpedantic` for GCC/Clang and `/W4` for MSVC, and remove as many warning suppressions as possible.

NVIDIA/cub#171 may be related, as it is another NVRTC issue.

# Deliverables

- [ ] `[1.12]` Enable `-Wpedantic` / `/W4` in our builds and fix the issues identified by these flags.
- [ ] `[1.12]` Audit the [current warning suppressions](https://github.com/NVIDIA/cub/blob/main/cmake/CubBuildCompilerTargets.cmake) and remove any that are hiding functional problems. Such issues should be fixed instead.

# Original Issue

I've repurposed this existing issue about anonymous structs to a larger purpose of improving C++ standard compliance across Thrust and CUB. The original issue was:

Title:  CUB uses anonymous structs which is disallowed in C++

For example see

https://github.com/NVIDIA/cub/blob/ea48955fe5814b2319f77a68bd7094f5fdbf1b08/cub/block/specializations/block_reduce_raking_commutative_only.cuh#L98

This is illegal in C++, see for example

https://stackoverflow.com/questions/2253878/why-does-c-disallow-anonymous-structs

This causes issues with nvrtc which is a strict C++ compiler.

allisonvacanti:
Thanks for pointing this out.

I don't see any gcc warning flags for this (other than `-Wpedantic`), but [clang can issue warn on these](https://clang.llvm.org/docs/DiagnosticsReference.html#wgnu-anonymous-struct). 

We should enable that and clean these up.

allisonvacanti:
I'm going to try fixing this by enabling `-Wpedantic`/`/W4` for our Thrust/CUB builds. This should hopefully prevent other non-standard extensions from creeping in and breaking NVRTC.

allisonvacanti:
Slipping to 1.12.0, since this won't be done in time for 1.11.0.

allisonvacanti:
Looks like using `-Wpedantic` with `nvcc` has issues. Thousands of non-suppressible warnings are emitted for the simplest translation units:

```
$ cat t.cu
int main() {}

[...]

Issue #221, opened 2020-10-19
NVBug: 3075796 [PyTorch] temp_storage_bytes overflows in InclusiveScan for size_cub value close to int32 max


allisonvacanti:
Reported by upstream.

We've created this code snippet to reproduce it in our current container:

```python
import torch
import torch.nn as nn
from torch.utils import cpp_extension

cuda_source = """
#include <cub/device/device_scan.cuh>

void my_fun(void)
{
    int size_cub = 2147483647-100000;
    auto self = torch::ones({size_cub});
    auto result = self.clone();
    size_t temp_storage_bytes = 0;
    cub::DeviceScan::InclusiveScan(nullptr, temp_storage_bytes, self.data_ptr<float>(), result.data_ptr<float>(), [] __host__ __device__ (const float& a, const float& b){return (b < a) ? b: a;}, size_cub);
    std::cout << "temp_storage_bytes " << temp_storage_bytes << std::endl;
    return;
}
"""

cpp_source = """
    void my_fun(void);
"""

module = torch.utils.cpp_extension.load_inline(
    name="cuda_test_extension",
    cpp_sources=cpp_source,
    cuda_sources=cuda_source,
    functions="my_fun",
    extra_cuda_cflags=["--extended-lambda"],
    verbose=True,
)

module.my_fun()

print('done')
```

Output for different values of `size_cub`:
```
2147483647
temp_storage_bytes 18446744073700604415

2147483647 - 10
temp_storage_bytes 18446744073700604415

2147483647 - 100
temp_storage_bytes 18446744073700604415

2147483647 - 1000
temp_storage_bytes 18446744073700604415

2147483647 - 10000
temp_storage_bytes 8948479

2147483647 - 100000
temp_storage_bytes 8947967
```

Based on this, it seems `temp_storage_bytes` overflows for a `size_cub` value between `[2147483647 - 10000, 2147483647 - 1000]`.

allisonvacanti:
@ptrblck Looks like this is caused by an overflow here: https://github.com/NVIDIA/cub/blob/main/cub/device/dispatch/dispatch_scan.cuh#L299

I'll fix this as part of #249, since that has some related fixes.

@brycelelbach [you were right](https://github.com/NVIDIA/cub/pull/249#discussion_r572352539) :)
(different location but same bug)

Issue #191, opened 2020-07-06
Deprecate TexRefInputIterator; underlying CUDA APIs are deprecated.


allisonvacanti:
These will need to be refactored out at some point.

PR #249, opened 2020-12-11
Enable more warning flags.

allisonvacanti:
Fixes NVIDIA/cub#228.

### DIFF
--- a/cmake/CubBuildCompilerTargets.cmake
+++ b/cmake/CubBuildCompilerTargets.cmake
@@ -11,20 +11,29 @@ function(cub_build_compiler_targets)
   set(cxx_compile_options)
 
   if ("MSVC" STREQUAL "${CMAKE_CXX_COMPILER_ID}")
-    # TODO Enable /Wall
+    append_option_if_available("/W4" cxx_compile_options)
+
     append_option_if_available("/WX" cxx_compile_options)
 
-    # Disabled loss-of-data conversion warnings.
-    # TODO Re-enable.
-    append_option_if_available("/wd4244" cxx_compile_options)
-    append_option_if_available("/wd4267" cxx_compile_options)
-
-    # Suppress numeric conversion-to-bool warnings.
-    # TODO Re-enable.
-    append_option_if_available("/wd4800" cxx_compile_options)
-
-    # Disable warning about applying unary operator- to unsigned type.
-    append_option_if_available("/wd4146" cxx_compile_options)
+    # Suppress overly-pedantic/unavoidable warnings brought in with /W4:
+    # C4324: structure was padded due to alignment specifier
+    append_option_if_available("/wd4324" cxx_compile_options)
+    # C4127: conditional expression is constant
+    # This can be fixed with `if constexpr` when available, but there's no way
+    # to silence these pre-C++17.
+    # TODO We should have per-dialect interface targets so we can leave these
+    # warnings enabled on C++17:
+    append_option_if_available("/wd4127" cxx_compile_options)
+    # C4505: unreferenced local function has been removed
+    # The CUDA `host_runtime.h` header emits this for
+    # `__cudaUnregisterBinaryUtil`.
+    append_option_if_available("/wd4505" cxx_compile_options)
+    # C4706: assignment within conditional expression
+    # MSVC doesn't provide an opt-out for this warning when the assignment is
+    # intentional. Clang will warn for these, but suppresses the warning when
+    # double-parentheses are used around the assignment. We'll let Clang catch
+    # unintentional assignments and suppress all such warnings on MSVC.
+    append_option_if_available("/wd4706" cxx_compile_options)
 
     # Some tests require /bigobj to fit everything into their object files:
     append_option_if_available("/bigobj" cxx_compile_options)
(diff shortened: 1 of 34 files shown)


cmake/CubBuildCompilerTargets.cmake
```diff
@@ -35,45 +44,34 @@ function(cub_build_compiler_targets)
...
-    append_option_if_available("-Wno-variadic-macros" cxx_compile_options)
+    append_option_if_available("-Wpointer-arith" cxx_compile_options)
+    append_option_if_available("-Wunused-local-typedef" cxx_compile_options)
+    append_option_if_available("-Wvla" cxx_compile_options)
+
+    # Disable GNU extensions (flag is clang only)
+    append_option_if_available("-Wgnu" cxx_compile_options)
+    # Calling a variadic macro with zero args is a GNU extension until C++20,
+    # but the THRUST_PP_ARITY macro is used with zero args. Need to see if this
+    # is a real problem worth fixing.
```
brycelelbach, line 55: It shouldn't be a problem as every compiler under the sun supports the extension, including NVRTC I believe.

cub/grid/grid_even_share.cuh
```diff
@@ -122,17 +122,18 @@ public:
...
-        this->block_offset          = num_items;    // Initialize past-the-end
-        this->block_end             = num_items;    // Initialize past-the-end
-        this->num_items             = num_items;
-        this->total_tiles           = (num_items + tile_items - 1) / tile_items;
-        this->grid_size             = CUB_MIN(total_tiles, max_grid_size);
+        this->block_offset          = num_items_;    // Initialize past-the-end
+        this->block_end             = num_items_;    // Initialize past-the-end
+        this->num_items             = num_items_;
+        this->total_tiles           = (num_items_ + tile_items - 1) / tile_items;
+        this->grid_size             = CUB_MIN(static_cast<int>(total_tiles), max_grid_size);
```
brycelelbach, line 133: This could overflow, because `total_tiles` is `OffsetT` which could be 64bit.

brycelelbach: This may be related to #221.

brycelelbach: I think I'd prefer us to either make sure all the static casts don't overflow, or move them to a separate PR.

cub/grid/grid_even_share.cuh
```diff
@@ -122,17 +122,18 @@ public:
...
         OffsetT avg_tiles_per_block = total_tiles / grid_size;
-        this->big_shares            = total_tiles - (avg_tiles_per_block * grid_size);        // leftover grains go to big blocks
+        // leftover grains go to big blocks
+        this->big_shares = static_cast<int>(total_tiles - (avg_tiles_per_block * grid_size));
```
brycelelbach, line 136: Is there any reason `big_share` can't be an `OffsetT`?

cub/iterator/tex_ref_input_iterator.cuh
```diff
@@ -42,6 +42,18 @@
...
+// See issue NVIDIA/cub#191.
+// Turn off deprecation warnings when compiling this file until then.
+#if CUB_HOST_COMPILER == CUB_HOST_COMPILER_MSVC
+#pragma warning(disable:4996)
+#elif CUB_HOST_COMPILER == CUB_HOST_COMPILER_GCC || \
+      CUB_HOST_COMPILER == CUB_HOST_COMPILER_CLANG
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
+
```
brycelelbach, line 56: I would deprecate or remove this entirely.

test/half.h
```diff
@@ -114,7 +114,7 @@ struct half_t
                 { /* normal */
                     ir |= ia >> (24 - 11);
                     ia = ia << (32 - (24 - 11));
-                    ir = ir + ((14 + shift) << 10);
+                    ir = static_cast<uint16_t>(ir + ((14 + shift) << 10));
```
brycelelbach, line 117: This is probably okay because it's just in the test code, but I'm pretty sure this is actually a bug due to potential overflow.

test/test_device_run_length_encode.cu
```diff
@@ -779,11 +779,11 @@ void TestSize(
...
         for (int i = 0; i < 10; ++i)
         {
-            unsigned int num_items;
-            RandomBits(num_items);
-            num_items = (unsigned int) ((double(num_items) * double(10000000)) / double(max_int));
-            num_items = CUB_MAX(1, num_items);
-            TestDispatch<T, OffsetT, LengthT>(num_items);
+            unsigned int num;
+            RandomBits(num);
+            num = (unsigned int) ((double(num) * double(10000000)) / double(max_int));
```
brycelelbach, line 784: Since we're changing this already, maybe `static_cast` here instead?

brycelelbach (approved):
This looks fine, I'd prefer to split the `static_cast` into a separate PR, or make sure they don't overflow, maybe add a comment to each explaining why they're fine.

allisonvacanti:
> You should re-run this with CUB in `-DDEBUG` debugging mode - you'll hit a LOT more of the assignment in conditional warning.

Updated #1175 with this recommendation.

cub/grid/grid_even_share.cuh
allisonvacanti: Fixed this and similar issues in the thrust/cub kernel dispatch code with a new `cub::DivideAndRoundUp` method that avoids these `(n + d - 1) / d` overflows.

cub/grid/grid_even_share.cuh
allisonvacanti: I took a closer look and if anything, `total_tiles` should be changed to a 32-bit int since it's used as a block dimension and those can't exceed INT32_MAX. The only usages of the variable assume it is a 32-bit int.

I'll change this so that `total_tiles` is appropriately sized, which will eliminate the need for this (and a few other) `static_cast`s.

cub/iterator/tex_ref_input_iterator.cuh
allisonvacanti: #191

test/test_device_run_length_encode.cu
allisonvacanti: Done.

test/test_device_scan.cu
```diff
@@ -914,11 +914,11 @@ void TestSize(
...
         for (int i = 0; i < 10; ++i)
         {
-            unsigned int num_items;
-            RandomBits(num_items);
-            num_items = (unsigned int) ((double(num_items) * double(10000000)) / double(max_int));
-            num_items = CUB_MAX(1, num_items);
-            TestOp<InputT>(num_items,  identity, initial_value);
+            unsigned int num;
+            RandomBits(num);
+            num = (unsigned int) ((double(num) * double(10000000)) / double(max_int));
```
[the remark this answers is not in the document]
allisonvacanti: Done.

allisonvacanti:
> This looks fine, I'd prefer to split the `static_cast` into a separate PR, or make sure they don't overflow, maybe add a comment to each explaining why they're fine.

After fixing #221, I reviewed the remaining usages and all are safe. Added comments to any non-obviously-safe ones.

Issue #250, opened 2020-12-12
Take C++ dialect out of the name of built CUB examples/tests


brycelelbach:
Always having the C++ dialect in the CUB test name (e.g. `cub.cpp14.test.device_select_unique.minimal`) is going to be problematic for testing NVC++ with CUB in the NVC++ test system. Can we fix this? Thrust doesn't do this when multi-config is off. I'd like that behavior.

I'd like to get the CUB tests turned on in NVC++'s test system ASAP so if we can prioritize this it would be great.

allisonvacanti:
I was actually planning to implement CUB's behavior in Thrust as part of NVIDIA/thrust#1333 -- keep the info in the test / example executable names and just have the one path through the build system to maintain. Configuring for a single config would still be possible (and default to cpp.cuda.cpp14), but the info would stay in the test / executable names.

Can you explain how this breaks nvc++builds? It sounds like I'll need to rethink my plans for thrust.

brycelelbach:
The NVC++ QA system wants to be able to pass through arbitrary flags to the compiler, which is how they manage configs for things like language dialect. Everything is just a flag for them.

We could just have an option to exclude it from the config name specifically for them.

PR #251, opened 2020-12-15
Add CUB_ENABLE_CPP_DIALECT_IN_NAMES CMake flag

allisonvacanti:
On by default. When disabled:
a) Emit configuration error if more than one dialect option is enabled, and
b) Remove dialect information from target / executable names.

Fixes NVIDIA/cub#250.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,14 @@ option(CUB_ENABLE_THOROUGH_TESTING "Build CUB thorough test variants." ON)
 option(CUB_ENABLE_BENCHMARK_TESTING  "Build CUB benchmark test variants." ON)
 option(CUB_ENABLE_MINIMAL_TESTING "Build CUB minimal test variants." ON)
 option(CUB_ENABLE_EXAMPLES "Build CUB examples." ON)
+# This is needed for NVCXX QA, which requires a static set of executable names.
+# Only a single dialect may be enabled when this is off.
+option(CUB_ENABLE_CPP_DIALECT_IN_NAMES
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -54,6 +54,9 @@ option(CUB_ENABLE_THOROUGH_TESTING "Build CUB thorough test variants." ON)
 option(CUB_ENABLE_BENCHMARK_TESTING  "Build CUB benchmark test variants." ON)
 option(CUB_ENABLE_MINIMAL_TESTING "Build CUB minimal test variants." ON)
 option(CUB_ENABLE_EXAMPLES "Build CUB examples." ON)
+option(CUB_NVCXX_QA_COMPAT "Use shorter names and force a single dialect." OFF)
```
brycelelbach, line 57: I'd prefer a name that describes what it does, not what it is. I'd suggest something like `CUB_ENABLE_CPP_DIALECT_IN_EXE_NAMES` (defaulted to `ON`). I don't think the name needs to capture the fact that you get a configuration error if you try to use it with multiple C++ dialects; that just seems like a natural consequence.

brycelelbach (changes requested):
Looks good, but I'm not thrilled with the name. I'd prefer something that describes what it does, not what it's used for.

allisonvacanti:
@brycelelbach Updated with requested changes.

run tests

brycelelbach (approved):
Approved, please just fix the comment I pointed out. Let me know when this is merged so I can add CUB tests to NVC++'s internal testing!

PR #254, opened 2020-12-24
Remove NPP dependency from `test_device_histogram.cu`. 

brycelelbach:
The code has been broken forever, we don't need it, and NVHPC doesn't have NPP, so it fails to build this test when `npp.h` is not found.


allisonvacanti:
I've already got this as part of #249 since it came up there, too. We can fix it here if you need it sooner, though.

allisonvacanti (approved):

PR #259, opened 2021-01-28
Remove some problematic inline annotations.

allisonvacanti:
These functions started producing invalid results in CUDA 11 under
certain circumstances (see issue NVIDIA/thrust#1371), and removing
these hints fixes the issue.

cc: @ArEsKay3
Fixes NVIDIA/thrust#1371


brycelelbach (approved):
There's no reason for this to have been `__forceinline__` in the first place (and generally `__forceinline__` is just bad).

brycelelbach:
This may have had something to do with the use of a static variable internally.

PR #263, opened 2021-02-10
Fix typo, s/Dipatch/Dispatch/g

allisonvacanti:


brycelelbach (approved):
There used to be a lot of these, I thought we got them all, but I guess not...

PR #272, opened 2021-03-17
Fix uses of CUB_PTX_ARCH in CUB tests to work with NVC++

dkolsen-pgi:
Using `CUB_PTX_ARCH` to conditionally compile code for host or device doesn't
work correctly with NVC++'s unified compilation model.  Change some of the
uses of `CUB_PTX_ARCH` in CUB test code to work with both NVCC and NVC++.
This fixes test_warp_reduce when compiled with NVC++.  The uses of
`CUB_PTX_ARCH` in the main CUB code were dealt with some time ago.

### DIFF
--- a/test/test_util.h
+++ b/test/test_util.h
@@ -537,45 +537,54 @@ __host__ __device__ __forceinline__ void InitValue(GenMode gen_mode, T &value, i
 {
     switch (gen_mode)
     {
-#if (CUB_PTX_ARCH == 0)
-    case RANDOM:
-        RandomBits(value);
-        break;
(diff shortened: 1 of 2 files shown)


dkolsen-pgi:
I have done some testing of this change with NVC++.  It needs to be tested with NVCC before merging.

allisonvacanti:
@dkolsen-pgi I'll be overwriting these changes with the `<nv/target>` refactor, which will hopefully land next week.

I'll probably just wait for that to land and save a few trips to our internal CI unless you need this sooner. Feel free to reopen if this is urgent.

PR #277, opened 2021-03-29
Fixed sanitizer warnings in RadixSortScanBinsKernel

canonizer:
Fixed sanitizer warnings in `RadixSortScanBinsKernel`.

- correct per-pass spine length
- correctly handling the last scan tile
- fixes warnings from https://github.com/NVIDIA/thrust/issues/1400

### DIFF
--- a/cub/device/dispatch/dispatch_radix_sort.cuh
+++ b/cub/device/dispatch/dispatch_radix_sort.cuh
@@ -167,6 +167,13 @@ __global__ void RadixSortScanBinsKernel(
         block_scan.template ConsumeTile<false, false>(block_offset, prefix_op);
         block_offset += AgentScanT::TILE_ITEMS;
     }
+
+    // Process the remaining partial tile (if any).
+    if (block_offset < num_counts)
+    {
+        block_scan.template ConsumeTile<false, true>(block_offset, prefix_op,
+                                                     num_counts - block_offset);
+    }
 }
 
 
@@ -1102,7 +1109,7 @@ struct DispatchRadixSort :
         const ValueT    *d_values_in,
         ValueT          *d_values_out,
         OffsetT         *d_spine,
-        int             spine_length,
+        int             /*spine_length*/,
         int             &current_bit,
         PassConfigT     &pass_config)
     {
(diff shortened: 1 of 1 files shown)


cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1116,6 +1123,10 @@ struct DispatchRadixSort :
                 pass_config.even_share.grid_size, pass_config.upsweep_config.block_threads, (long long) stream,
                 pass_config.upsweep_config.items_per_thread, pass_config.upsweep_config.sm_occupancy, current_bit, pass_bits);
 
+            // Zero the spine.
+            //if (CubDebug(error = cudaMemsetAsync(d_spine, 0, sizeof(OffsetT) * spine_length, stream))) break;
```
allisonvacanti, line 1127: Was this intentional? We should uncomment or remove these lines.

allisonvacanti (changes requested):
Question about the implementation.

Once the patch is finalized we should do a quick check to see if this has any significant perf impact.

allisonvacanti:
@canonizer I'd like to get this in sometime this week to land in 1.13. Should I just remove the dead code mentioned in my review? Is there anything else that this needs to be ready?

cub/device/dispatch/dispatch_radix_sort.cuh
canonizer: This was something I tried to do. Feel free to remove it.

allisonvacanti:
DVS CL: 30066471
gpuCI: NVIDIA/thrust#1460

allisonvacanti:
Fixed a new warning.

DVS CL: 30073602
gpuCI: NVIDIA/thrust#1460

jlebar:
Found this PR from the release notes for the 1.13 RC.

What is the practical effect of this change?  Without it, we do OOB memory reads (and writes?), which might crash depending on page sizes?

canonizer:
I don't think there's any practical effect, apart from fixing the sanitizer warning.

There has been no out-of-bounds reads. Rather, uninitialized memory was fed into a scan, which caused them to be read, which triggered a warning. However, the results that were derived from uninitialized memory locations weren't used anywhere, so the original warning was benign.

Issue #282, opened 2021-04-08
Document that cub::DeviceRadixSort and cub::BlockRadixSort are stable


zasdfgbnm:
There is nothing in the documentation https://nvlabs.github.io/cub/structcub_1_1_device_radix_sort.html saying whether it is stable sort or not. It seems to be stable. If so, it would be great if this fact is mentioned in the documentation.

allisonvacanti:
Agreed. For the record, it is stable.

We're revamping the documentation page right now, we should update the docs to mention this soon.

himanshu007-creator:
HI! i would like to work on this issue!

allisonvacanti:
Sounds great! I'll assign it to you, just make a pull request with the updated docs when you have it ready.

Thanks!

himanshu007-creator:
@allisonvacanti , can you help me locate the file please?

allisonvacanti:
@himanshu007-creator Sure -- you'll just need to update this line:

https://github.com/NVIDIA/cub/blob/main/cub/device/device_radix_sort.cuh#L69

Before:

```
For floating-point types...
```

After:

```
This radix sort implementation is stable, and for floating-point types...
```

zasdfgbnm:
Could we also add documentation for `cub::BlockRadixSort` and `cub::DeviceSegmentedRadixSort`?

zasdfgbnm:
ping @allisonvacanti Are `cub::BlockRadixSort` and `cub::DeviceSegmentedRadixSort` stable?

allisonvacanti:
Yes, the block radix sorts are stable. We should update the docs for this -- @himanshu007-creator could you update #284 and add the same docstring to the `cub::BlockRadixSort::Sort*` functions too?

https://github.com/NVIDIA/cub/blob/main/cub/block/block_radix_sort.cuh#L477

All of the functions following this one should be documented as stable.

zasdfgbnm:
How about `cub::DeviceSegmentedRadixSort`? I assume it is stable as well.

allisonvacanti:
Yes, all of the radix sorts are inherently stable -- let me know if you spot any others that I've missed, but I think that's all of them 😅

@himanshu007-creator the Sort* functions in https://github.com/NVIDIA/cub/blob/main/cub/device/device_segmented_radix_sort.cuh should also have the stability docstring in #284.

PR #284, opened 2021-04-14
Update device_radix_sort.cuh

himanshu007-creator:
#282

### DIFF
--- a/cub/device/device_radix_sort.cuh
+++ b/cub/device/device_radix_sort.cuh
@@ -67,6 +67,7 @@ namespace cub {
  * method can only be applied to unsigned integral types, DeviceRadixSort
  * is able to sort signed and floating-point types via simple bit-wise transformations
  * that ensure lexicographic key ordering. For floating-point types -0.0 and +0.0 are
+ * cub::DeviceRadixSort is stable
  * considered equal and appear in the result in the same order as they appear in
  * the input.
  *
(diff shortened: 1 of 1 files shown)


cub/device/device_radix_sort.cuh
```diff
@@ -66,7 +66,7 @@ namespace cub {
  * half-precision floating-point type.  Although the direct radix sorting
  * method can only be applied to unsigned integral types, DeviceRadixSort
  * is able to sort signed and floating-point types via simple bit-wise transformations
- * that ensure lexicographic key ordering. For floating-point types -0.0 and +0.0 are
+ * that ensure lexicographic key ordering. DeviceRadixSort is stable For floating-point types -0.0 and +0.0 are
```
allisonvacanti, line 69: This needs to be blended into the text a bit more. It reads as "Device radix sort is stable for floating-point types", which is true, but the results are stable for non-floating-point types, too,

Can we switch this to "DeviceRadixSort is stable, and for floating-point types..." (or just make them separate sentences)?

allisonvacanti (changes requested):

cub/device/device_radix_sort.cuh
himanshu007-creator: sure

allisonvacanti (approved):

allisonvacanti:
Oops, I missed your update. This LGTM, I'll merge it soon -- thanks again!

BTW, don't feel obligated, but you may be interested in #289 :)

allisonvacanti:
Replaced by PR #307, which adds the same info to the other radix sort algorithms.

PR #291, opened 2021-04-28
Allow segmented problems to have different types for offset iterators.

psvvsp:
Modified segmented reduce, segmented radix sort and their tests.

### DIFF
--- a/test/test_device_radix_sort.cu
+++ b/test/test_device_radix_sort.cu
@@ -45,6 +45,7 @@
 #include <cub/util_math.cuh>
 #include <cub/device/device_radix_sort.cuh>
 #include <cub/device/device_segmented_radix_sort.cuh>
+#include <cub/iterator/transform_input_iterator.cuh>
 
 #include "test_util.h"
 
@@ -85,7 +86,7 @@ enum Backend
 /**
  * Dispatch to CUB sorting entrypoint (specialized for ascending)
  */
-template <typename KeyT, typename ValueT>
+template <typename KeyT, typename ValueT, typename BeginOffsetIteratorT, typename EndOffsetIteratorT>
 CUB_RUNTIME_FUNCTION
 __forceinline__
 cudaError_t Dispatch(
@@ -101,7 +102,8 @@ cudaError_t Dispatch(
     DoubleBuffer<ValueT>    &d_values,
     int                     num_items,
     int                     /*num_segments*/,
-    const int               */*d_segment_offsets*/,
+    BeginOffsetIteratorT    /*d_segment_begin_offsets*/,
+    EndOffsetIteratorT      /*d_segment_end_offsets*/,
     int                     begin_bit,
     int                     end_bit,
     cudaStream_t            stream,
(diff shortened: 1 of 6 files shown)


test/test_device_radix_sort.cu
```diff
@@ -982,16 +1002,67 @@ void TestBits(
...
+
+    // Test with segment iterator
+    typedef CastOp<int> IdentityOpT;
+    IdentityOpT identity_op;
+    TransformInputIterator<int, IdentityOpT, int*, int> d_segment_offsets_itr(d_segment_offsets, identity_op);
+
+    TestBits(h_keys, num_items, num_segments, h_segment_offsets, d_segment_offsets_itr, d_segment_offsets_itr + 1);
+
+    // Test with thrust transform iterators
+    auto d_segment_begin_offsets_itr = thrust::make_transform_iterator(d_segment_offsets, transform_functor1<int>());
```
allisonvacanti, line 1062: This adds an include for `cub/iterator/transform_input_iterator.cuh`, but uses Thust's transform iterators. Please update this to use the CUB iterators in the implementation.

test/test_device_reduce.cu
```diff
@@ -1106,8 +1159,20 @@ void TestByBackend(
             d_segment_offsets,
             identity_op);
 
-        TestByOp<CUB_SEGMENTED, OutputT>(
-            h_in, d_in, num_items, num_segments, h_segment_offsets_itr, d_segment_offsets_itr);
+        TestByOp<CUB_SEGMENTED, OutputT>(h_in, d_in, num_items, num_segments,
+            h_segment_offsets_itr, h_segment_offsets_itr + 1, d_segment_offsets_itr, d_segment_offsets_itr + 1);
+
+        // Test with thrust transform iterators
```
allisonvacanti, line 1165: Same as above -- let's use the CUB abstractions if possible.

allisonvacanti (changes requested):
Thanks for the patch, this looks great!

One small change needed in the tests -- the transform iterator header and implementation are mismatched. Once that's fixed I'll kick off testing for this.

test/test_device_radix_sort.cu
psvvsp: I used Thrust's iterators because they were mentioned in the problem description. Are you completely sure the Thrust's iterators should be replaced with the CUB ones?

allisonvacanti: Yes -- as long as the offset iterators are different types, the test will be sufficient for the feature. It doesn't matter which transform iterator implementation is used.

For context, right now there is a circular dependency between Thrust and CUB that I'm trying to break. Thrust must depend on CUB for its algorithm implementations, but CUB doesn't have to depend on Thrust, so I'm trying to avoid adding any new Thrust usages in CUB.

allisonvacanti:
Looks good -- thanks again for this! I know these large, tedious refactorings aren't fun, and I appreciate your effort here :)

I'll kick off testing today, we should be able to get this into the next release.

allisonvacanti:
run tests

allisonvacanti (approved):

allisonvacanti:
gpuCI: NVIDIA/thrust#1422
DVS: 29929847

psvvsp:
Will you write me if the tests fail? I'm ready to fix bugs if you provide me with build configuration.

allisonvacanti:
Sure! I'll let you know how it goes.

Issue #294, opened 2021-05-03
[RFE] Allow `init_value` to be a device pointer in `cub::DeviceScan::ExclusiveScan`


zasdfgbnm:
per title

allisonvacanti:
I assume this is to support a usecase where `init_value` isn't yet computed when launching the scan kernel? That would definitely be nice to have.

zasdfgbnm:
@allisonvacanti Yes, correct.

allisonvacanti:
Gotcha. I'm not sure when something like this would happen, it would need to go into a new overload and require some invasive changes to the implementation, but I'll keep it in mind when as we start updating algorithm implementations in the coming months.

zasdfgbnm:
The use case in PyTorch is to manually split tensors larger than 2^31 elements into small tensors. In this use case, we will run the exclusive scan on the first tensor, and use the last element in the first tensor as the `init_value` of the second tensor.

ngimel:
Turns out we can use custom type for `InitValueT` to do what we want 
```
template <typename T>
struct init_val {
    __host__ __device__ init_val(T* init_ptr){
        init_p = init_ptr;
    }
    __host__ __device__ operator T() const {return *init_p;}
    T * init_p;
};
```
As far as I could tell inside the kernel `InitValueT` is implicitly converted to `OutputValueT` so the above struct makes it possible.

allisonvacanti:
Thanks for pointing this out @ngimel! I think that's a much cleaner solution that works well within the design of the library.

Closing this issue. The suggestion above is preferable to adding a new overload for this usecase.

ngimel:
@allisonvacanti this solution relies on cub internal implementation, i.e., that the first and last thing it does with `init_value` is implicitly or explicitly convert it to OutputValueT. It seems to work now, but cub could choose to do something internally that would break the above code, so it would be good to explicitly list the requirements that you have for `InitValueT`.
And yes, thanks for making `InitValueT` and `OutputValueT` different template types, that was a good design decision that makes cub much more flexible!

allisonvacanti:
Oops, yeah 😅 Looks like that change may have broken your usecase...

Pinging @brycelelbach for visibility.

zasdfgbnm:
Can we add as special class `cub::InitValueFromDevicePointer`, something like:
```C++
template<typename InitValueT>
struct InitValueFromDevicePointer {
    InitValueT *ptr;
[...]

PR #305, opened 2021-05-12
[Discussion] Support device pointer for initial value

zasdfgbnm:


allisonvacanti:
I think this idea is very useful, and could actually be more general than just `InitialValue` usecases. I haven't reviewed the implementation too closely yet, but some initial impressions/thoughts:

1. `InitValueFromDevicePointer` is too narrow of a name. This utility behaves similarly to a simplified `std::future`, so maybe something like `cub::SimpleFutureValue` would be more suitable?
2. The current implementation lacks a lot of the safety features of a traditional `future` -- in particular, there's no way to check if the value has actually been set. This is why I'm thinking of calling it a `SimpleFutureValue` instead of just `FutureValue`.
3. We /could/ flesh out the API into a more complete `future` implementation, but I'm not sure if this is really valuable here. The usecase of "I'm computing a value at this address in a previous stream-ordered kernel" wouldn't benefit from the additional safety checks since synchronization is handled externally. We'd just need to carefully document this.
4. We should look through the other device algorithms and see where else this may be useful. We'll need to explicitly document which algorithms support using this utility.

@jrhemstad @brycelelbach @griwes @dkolsen-pgi @dumerrill @canonizer -- Any thoughts on this? Alternative names, possible issues, etc? There's some more context in #294.

zasdfgbnm:
I think it can be even more general than just "a future on stream order wrapping a device pointer". We can make it a lazy value, something like:
```C++
ptr = ....;
cub::LazyValue<InitialValue>([=]()__device__ { return *ptr; });
```

zasdfgbnm:
@elstehle Is there a reliable way to test if an object is an iterator? Your suggestion sounds like a good design, but it might be tricky to implement.

Also, I don't have a strong preference on how the method should be called, and how it should behave. The design suggestion from @allisonvacanti and @elstehle both look good to me. I am OK with whichever decision we made. But I do hope we don't spend too much time making decisions so that I can start using this feature early.

allisonvacanti:
> I do hope we don't spend too much time making decisions so that I can start using this feature early.

Unfortunately my time is very limited right now, so it may be a little while before I can spend too many cycles on this. The good news is that I'm getting some more help working on CUB, so I should have more time for this sort of stuff in the near future.

I'll keep this on my radar, but feel free to remind me if we don't address it in the next couple of weeks. I do see this as an important usecase, but I want to make sure we get it right.

allisonvacanti:
I was thinking about this some more over the weekend. I think we can get this into 1.14 -- here's what I'd like to see:

- Name the wrapper utility `cub::FutureValue<T>`.
  - It just holds a device pointer.
  - No synchronization or safety checks; user must ensure that the result will be ready before use via external synchronization or stream-ordering dependencies.
  - Document these sharp edges well.
- Consistency: All public `Device` level entry points that currently take immediate values instead of a pointer should have `cub::FutureValue<T>` overloads.
- Low-overhead: We must not instantiate multiple kernels / implementations depending on whether a `FutureValue` or immediate value is provided.
  - The distinction should only exist in the public API layer.
  - This could be accomplished by having a wrapper that toggles between loading from a `FutureValue` or a stored value. This would be constructed by the public `Device` level functions, and passed into the Dispatch layer.

The wrapper in the last point could be as simple as:

```
template <typename T>
struct InputValue
{
  bool is_future;
  FutureValue<T> future_value;
  T immediate_value;
  // Plus API to construct (host/device) and access value (device-only)
};
```

This would be passed from the `Device` entry point down to the kernel, where the actual access happens.

I think this would address my concerns about this feature. We can start with reduce to work out any issues before porting the other algorithms in case there are surprises.

Does this sound reasonable?

allisonvacanti:
Just realized I'd neglected @elstehle suggestion for a more general iterator in the above. That complicates the implementation, but should be doable.

```
template <typename T, typename IterT = T*>
struct FutureValue
{
  using value_type = T;
  using iterator_type = IterT;
};

template <typename T, typename IterT = T*>
struct InputValue
{
  bool is_future;
  FutureValue<T, IterT> future_value;
  T immediate_value;
};
```

We would incur extra instantiation penalties for `IterT != T*`, but that's unavoidable, and the common case of a pointer can still be rolled into the default instantiation.

zasdfgbnm:
Hi @allisonvacanti, sorry for the late reply. This design looks good to me, I will ask if I have further questions. What is the timeline for 1.14 and 1.15? I am recently busy and I will schedule myself to try to catch the deadline of accepting PRs of 1.14 or 1.15 for this.

allisonvacanti:
@zasdfgbnm The milestones are tracked here: https://github.com/NVIDIA/cub/milestones. Currently 1.14 features need to be merged by 8/18, and 1.15 should be complete by 10/18.

Are you planning to handle the implementation for this?

zasdfgbnm:
@allisonvacanti Yes, I will. But I am not sure if it will be 1.14 or 1.15.

allisonvacanti:
Sounds good, I'll leave the milestone blank for now. Thanks!

zasdfgbnm:
Please see https://github.com/NVIDIA/thrust/pull/1519 for thrust change.

zasdfgbnm:
@allisonvacanti This should be ready for review

cub/device/dispatch/dispatch_scan.cuh
```diff
@@ -96,7 +95,7 @@ template <
     typename            OutputIteratorT,    ///< Random-access output iterator type for writing scan outputs \iterator
     typename            ScanTileStateT,     ///< Tile status interface type
     typename            ScanOpT,            ///< Binary scan functor type having member <tt>T operator()(const T &a, const T &b)</tt>
-    typename            InitValueT,         ///< Initial value to seed the exclusive scan (cub::NullType for inclusive scans)
+    typename            InputValueT,        ///< Initial value to seed the exclusive scan (cub::NullType for inclusive scans)
```
allisonvacanti, line 98: I wouldn't change the name of this -- I think `InitValueT` better describes what this represents.

cub/util_type.cuh
```diff
@@ -273,9 +274,47 @@ struct NullType
...
 {
-   enum {VALUE = A};
+    enum {VALUE = A};
 };
 
+/**
+ * \brief Allows algorithms that take a value as input to take a future value that is not computed yet at launch time.
+ *
+ * Note that it is user's responsibility to ensure that the result will be ready before use via external synchronization
+ * or stream-ordering dependencies.
```
allisonvacanti, line 284: Can you add an example `@code`/`@endcode` block showing how this is used?

cub/util_type.cuh
```diff
@@ -273,9 +274,47 @@ struct NullType
...
+    using value_type = T;
+    using iterator_type = IterT;
+    IterT iter;
+    __host__ __device__ __forceinline__ FutureValue(IterT iter):iter(iter) {}
+    __host__ __device__ __forceinline__ operator T() {
+        return *iter;
+    }
+};
+
+template <typename T, typename IterT = T*>
```
allisonvacanti, line 298: I didn't think of this in our initial discussion, but let's move this into the `cub::detail::` namespace, since users shouldn't directly work with it. It will also need a documentation string explaining why it exists.

cub/util_type.cuh
```diff
@@ -273,9 +274,47 @@ struct NullType
...
+ * Note that it is user's responsibility to ensure that the result will be ready before use via external synchronization
+ * or stream-ordering dependencies.
+ */
+template <typename T, typename IterT = T*>
+struct FutureValue
+{
+    using value_type = T;
+    using iterator_type = IterT;
+    IterT iter;
+    __host__ __device__ __forceinline__ FutureValue(IterT iter):iter(iter) {}
```
allisonvacanti, line 292: - This constructor should be marked `explicit`.
- Use a different name than "iter" for either the argument or member to avoid shadowing warnings. I'd suggest renaming the member variable `m_iter`.
- Make the internal IterT member variable private, since users shouldn't directly interact with it.

cub/util_type.cuh
```diff
@@ -273,9 +274,47 @@ struct NullType
...
+        FutureValue<T, IterT> future_value;
+        T immediate_value;
+    };
+    __host__ __device__ __forceinline__ operator T() {
+        if (is_future) {
+            return future_value;
+        }
+        return immediate_value;
+    }
+    __host__ __device__ __forceinline__ InputValue(T immediate_value): is_future(false), immediate_value(immediate_value) {}
```
allisonvacanti, line 315: Same notes as above:
- These constructors should be marked `explicit`.
- Use a different name for either the argument or member to avoid shadowing warnings. I'd suggest prefixing the member variables with `m_` since we're considering adopting this style convention "soon".
- Make the internal member variable + union private, since users shouldn't directly interact with them.

allisonvacanti (commented):
This looks great! I pointed out some minor issues in comments, but I'm very happy with how this is turning out :)

@jrhemstad @elstehle @senior-zero What do you think?

cub/util_type.cuh
zasdfgbnm: fixed

cub/util_type.cuh
zasdfgbnm: added

cub/device/dispatch/dispatch_scan.cuh
zasdfgbnm: fixed

cub/util_type.cuh
zasdfgbnm: fixed

cub/util_type.cuh
zasdfgbnm: fixed

zasdfgbnm (commented):
@allisonvacanti I have resolved all your review comments.

allisonvacanti:
Thanks! I'm happy with this implementation, thanks for your patience and effort on it!

One last thing -- this will need to be rebased on the current `main` branch before it will merge cleanly. Can you take care of that, and also squash this branch down to a single commit?

Once that's taken care of I'll start testing.

zasdfgbnm:
done

PR #306, opened 2021-05-12
Add BFloat16 support for radix sort

zasdfgbnm:

Issue #293, opened 2021-04-28
Document radix sort bitwise transformations on signed and floating-point types


zasdfgbnm:
The highest bit in signed integers is the sign bit. If I use `end_bit` < `sizeof(KeyT)*8`, cub will not consider sign bits. And I suggest cub to clearly document this fact.

The confusing part of this problem is, for example, if I am using `KeyT = int64_t` to sort `[0, 1, 2, 3]`, I can use `end_bit=2` to speed up the sorting (PyTorch uses `end_bit` to speedup sorting in many places). But if the keys are `[0, 1, -1]` and `end_bit=2`, I don't know whether cub will use some trick to make sure `-1 < 0`, or will just do the same bit transformation as if `end_bit=32`. The experiment shows me that it is the latter, but I prefer this to be documented so that we know this is defined behavior.

A C++ program to demonstrate this problem is:

```CUDA
#include <iostream>
#include <cub/cub.cuh>
#include <cmath>

constexpr size_t N = 10;
using type = int64_t;
using value_type = long;

template<typename T>
void print(T array[], size_t N) {
    for(size_t i = 0; i < N; i++) {
        std::cout << float(array[i]) << ", ";
    }
    std::cout << std::endl;
}

int main() {
    type *h1, *h2, *d1, *d2;
    value_type *d3, *d4;
    void *tmp;
    h1 = new type[N];
    h2 = new type[N];
    for(size_t i = 0; i < N; i++) {
        h1[i] = -i;
    }
    print(h1, N);

    cudaMalloc(&d1, sizeof(type) * N);
    cudaMalloc(&d2, sizeof(type) * N);
    cudaMalloc(&d3, sizeof(value_type) * N);
    cudaMalloc(&d4, sizeof(value_type) * N);
    cudaMemcpy(d1, h1, sizeof(type) * N, cudaMemcpyDefault);
    size_t temp_storage_bytes;
    cub::DeviceRadixSort::SortPairs(nullptr, temp_storage_bytes, d1, d2, d3, d4, N, 0, 10);
    cudaMalloc(&tmp, temp_storage_bytes);
    cudaDeviceSynchronize();
    cub::DeviceRadixSort::SortPairs(tmp, temp_storage_bytes, d1, d2, d3, d4, N, 0, 10);
    cudaDeviceSynchronize();
    cudaMemcpy(h2, d2, sizeof(type) * N, cudaMemcpyDefault);
    print(h2, N);
}
```

From the above program, we can see that
```
[...]

PR #307, opened 2021-05-12
Radix sort stability docs

### LINKED ISSUE #282: Document that cub::DeviceRadixSort and cub::BlockRadixSort are stable

allisonvacanti:
Fixes #282, extends and replaces #284.

### DIFF
--- a/cub/block/block_radix_sort.cuh
+++ b/cub/block/block_radix_sort.cuh
@@ -79,6 +79,13 @@ namespace cub {
  * (<tt>unsigned char</tt>, \p int, \p double, etc.) as well as CUDA's \p __half
  * half-precision floating-point type.
  *
+ * \par Floating-Point Special Cases
+ *
+ * - Positive and negative zeros are considered equivalent, and will be treated
+ *   as such in the output.
+ * - No special handling is implemented for NaN values; these are sorted
+ *   according to their bit representations after any transformations.
+ *
  * \par Bitwise Key Transformations
  * Although the direct radix sorting method can only be applied to unsigned
  * integral types, BlockRadixSort is able to sort signed and floating-point
--- a/cub/block/block_radix_sort.cuh
+++ b/cub/block/block_radix_sort.cuh
@@ -64,23 +64,53 @@ namespace cub {
  * \tparam PTX_ARCH             <b>[optional]</b> \ptxversion
  *
  * \par Overview
- * - The [<em>radix sorting method</em>](http://en.wikipedia.org/wiki/Radix_sort) arranges
- *   items into ascending order.  It relies upon a positional representation for
- *   keys, i.e., each key is comprised of an ordered sequence of symbols (e.g., digits,
- *   characters, etc.) specified from least-significant to most-significant.  For a
- *   given input sequence of keys and a set of rules specifying a total ordering
- *   of the symbolic alphabet, the radix sorting method produces a lexicographic
- *   ordering of those keys.
- * - BlockRadixSort can sort all of the built-in C++ numeric primitive types
- *   (<tt>unsigned char</tt>, \p int, \p double, etc.) as well as CUDA's \p __half
- *   half-precision floating-point type. Within each key, the implementation treats fixed-length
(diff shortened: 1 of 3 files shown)


cub/device/device_radix_sort.cuh
```diff
@@ -66,9 +66,9 @@ namespace cub {
  * half-precision floating-point type.  Although the direct radix sorting
  * method can only be applied to unsigned integral types, DeviceRadixSort
  * is able to sort signed and floating-point types via simple bit-wise transformations
- * that ensure lexicographic key ordering. For floating-point types -0.0 and +0.0 are
- * considered equal and appear in the result in the same order as they appear in
- * the input.
+ * that ensure lexicographic key ordering. DeviceRadixSort is stable. For
+ * floating-point types -0.0 and +0.0 are considered equal and appear in the
+ * result in the same order as they appear in the input.
```
zasdfgbnm, line 71: Can we also document that cub does not support in-place operations, so input and output pointers can not have memory overlap?

allisonvacanti: Absolutely, I'll add that before this goes in. 👍

allisonvacanti:
@zasdfgbnm I added some more documentation to the sort algorithms in the last push, including #293. Can you review these changes to make sure they provide enough information for your needs?

cub/block/block_radix_sort.cuh
```diff
@@ -64,23 +64,53 @@ namespace cub {
...
+ * given input sequence of keys and a set of rules specifying a total ordering
+ * of the symbolic alphabet, the radix sorting method produces a lexicographic
+ * ordering of those keys.
+ *
+ * \rowmajor
+ *
+ * \par Supported Types
+ * BlockRadixSort can sort all of the built-in C++ numeric primitive types
+ * (<tt>unsigned char</tt>, \p int, \p double, etc.) as well as CUDA's \p __half
+ * half-precision floating-point type.
```
zasdfgbnm, line 80: this might need to add `__nv_bfloat16`

cub/block/block_radix_sort.cuh
```diff
@@ -64,23 +64,53 @@ namespace cub {
...
+ *
+ * \par No Descending Sort Transformations
+ * Unlike `DeviceRadixSort`, `BlockRadixSort` does not invert the input key bits
+ * when performing a descending sort. Instead, it has special logic to reverse
+ * the order of the keys while sorting.
+ *
+ * \par Stability
+ * BlockRadixSort is stable. For floating-point types -0.0 and +0.0
+ * are considered equal and appear in the result in the same order as they
+ * appear in the input.
```
zasdfgbnm, line 112: How about adding something about the behavior of nan?

cub/device/device_segmented_radix_sort.cuh
```diff
@@ -66,9 +66,9 @@ namespace cub {
  * half-precision floating-point type.  Although the direct radix sorting
  * method can only be applied to unsigned integral types, DeviceSegmentedRadixSort
  * is able to sort signed and floating-point types via simple bit-wise transformations
- * that ensure lexicographic key ordering. For floating-point types -0.0 and +0.0 are
- * considered equal and appear in the result in the same order as they appear in
- * the input.
+ * that ensure lexicographic key ordering. DeviceSegmentedRadixSort is stable. For
+ * floating-point types -0.0 and +0.0 are considered equal and appear in the
+ * result in the same order as they appear in the input.
```
zasdfgbnm, line 71: Same here:
How about adding something about the behavior of nan?

cub/block/block_radix_sort.cuh
allisonvacanti: I have updated #306 to mention bfloat16. I'll update that PR after this goes in.

cub/block/block_radix_sort.cuh
allisonvacanti: 👍🏻

Issue #264, opened 2021-02-22
Potential Bug in the Down function of  cub/cub/block/block_shuffle.cuh 


sankarmanoj:
The [Down](https://github.com/NVIDIA/cub/blob/main/cub/block/block_shuffle.cuh#L264) function looks like a copy of the `Up` function, without any changes. 

It should look something like this instead:
```
    template <int ITEMS_PER_THREAD>
    __device__ __forceinline__ void Down(
        T (&input)[ITEMS_PER_THREAD],
        T (&next)[ITEMS_PER_THREAD])  
    {
        temp_storage[linear_tid].next = input[0];

        CTA_SYNC();

        #pragma unroll
        for (int ITEM = 0; ITEM < (ITEMS_PER_THREAD - 1); ++ITEM)
            next[ITEM] = input[ITEM - 1];

        if (linear_tid <(BLOCK_THREADS -1))
            next[0] = temp_storage[linear_tid + 1].next;
    }
```

Also, the Up function was called where the Down function should have been called in cub/cub/block/block_shuffle.cuh:295

allisonvacanti:
Thanks for letting us know -- this definitely needs to be fixed.

Issue #268, opened 2021-03-04
Temporary Storage Access in cub/cub/block/block_shuffle.cuh 


sankarmanoj:
_TempStorage was declared as 
```
    struct _TempStorage
    {
        T prev[BLOCK_THREADS];
        T next[BLOCK_THREADS];
    };

    _TempStorage &temp_storage;
```
So, access to a element in `temp_storage` should look like `temp_storage.prev[linear_tid]`. 
However, all the [accesses](https://github.com/NVIDIA/cub/blob/main/cub/block/block_shuffle.cuh#L174) in the file are like `temp_storage[linear_tid].prev`.

sankarmanoj:
I don't think you have any test cases that cover BlockShuffle. In fact, BlockAdjacentDifference, BlockRadixRank, & BlockDiscontinuity are also not covered in the CUB test cases. 

I added BlockShuffle to an existing BlockScan test case, by adding the following lines [here](https://github.com/NVIDIA/cub/blob/7cdf6dfc7ed60e4f44d025b84ea8260755f298e6/test/test_block_scan.cu#L416)
```
    typedef BlockShuffle<T, BLOCK_DIM_X, BLOCK_DIM_Y, BLOCK_DIM_Z> BlockShuffleT;
    __shared__ typename BlockShuffleT::TempStorage shuffle_temp_storage;
    BlockShuffleT                          block_shuffle(shuffle_temp_storage);
    block_shuffle.Offset(*d_in,*d_out);
```
which caused a compilation error.

PR #311, opened 2021-05-26
Fix block shuffle

senior-zero:
In order to fix [the issue](https://github.com/NVIDIA/cub/issues/264) with `BlockShuffle::Down` function I've written tests for `BlockShuffle` . It turned out that it was broken because of the wrong `_TempStorage` type which also led to extra shared memory usage.

### DIFF
--- a/test/test_block_shuffle.cu
+++ b/test/test_block_shuffle.cu
@@ -1,0 +1,369 @@
+/******************************************************************************
+ * Copyright (c) 2011-2021, NVIDIA CORPORATION. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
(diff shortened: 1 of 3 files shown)


senior-zero:
This PR also fixes [this issue](https://github.com/NVIDIA/cub/issues/268).

test/test_block_shuffle.cu
```diff
@@ -0,0 +1,358 @@
...
+
+    data_type *d_data = nullptr;
+    CubDebugExit(g_allocator.DeviceAllocate((void**)&d_data, sizeof(data_type) * num_items));
+
+    std::unique_ptr<data_type[]> h_data(new data_type[num_items]);
+
+    Test<data_type, items_per_thread, threads_in_block, UpTest>(num_items, d_data, h_data.get());
+    Test<data_type, items_per_thread, threads_in_block, DownTest>(num_items, d_data, h_data.get());
+
+    SingleItemTestHelper<data_type, items_per_thread, threads_in_block, OffsetUpTest>().run(num_items,
```
allisonvacanti, line 319: Can you add a comment explaining what this helper is used for? It looks like this becomes a no-op when `items_per_thread` != 1, but this is only invoked (in main) with `items_per_thread` == 1.

We should probably add some test cases where `items_per_thread` > 1.

allisonvacanti (changes requested):

allisonvacanti:
DVS CL: 30027318
gpuCI: NVIDIA/thrust#1451

allisonvacanti (approved):

PR #316, opened 2021-06-02
Remove WARP_TIME_SLICING template param desc.

Peter9606:
WARP_TIME_SLICING is contained in ALGORITHM and shouldn't be listed as a standalone template parameter.

Signed-off-by: Peter Han <fujun.han@iluvatar.ai>


allisonvacanti:
LGTM, this change is correct. Thanks for the patch!

PR #320, opened 2021-06-09
Deprecate TexRefInputIterator.

allisonvacanti:
This class uses deprecated CUDART APIs. The improved replacement, `TexObjInputIterator`, should be used instead. I'd like to get this in for 1.13.0 so folks get as much notice as possible.

Also updated the deprecation mechanism to provide an opt-out, cmake options, and C++14 syntax. Our internal build will opt-out of deprecation warnings so we can continue to maintain and test deprecated APIs.


allisonvacanti:
DVS CL: 30058645
gpuCI: NVIDIA/thrust#1459

cub/iterator/tex_ref_input_iterator.cuh
```diff
@@ -79,8 +67,21 @@ namespace {
...
+// See issue NVIDIA/cub#191.
+// Turn off deprecation warnings when compiling class implementation in favor
+// of deprecating TexRefInputIterator instead.
+#if CUB_HOST_COMPILER == CUB_HOST_COMPILER_MSVC
+#pragma warning(disable:4996)
+#elif CUB_HOST_COMPILER == CUB_HOST_COMPILER_GCC || \
+      CUB_HOST_COMPILER == CUB_HOST_COMPILER_CLANG
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
```
brycelelbach, line 83: We do this so that we don't trigger deprecation warnings in our own code, only when users use the deprecated facility, correct?

cub/iterator/tex_ref_input_iterator.cuh
```diff
@@ -79,8 +67,21 @@ namespace {
...
 /// Global texture reference specialized by type
 template <typename T>
-struct IteratorTexRef
+struct CUB_DEPRECATED IteratorTexRef
 {
+
+// This class uses the deprecated cudaBindTexture / cudaUnbindTexture APIs.
+// See issue NVIDIA/cub#191.
+// Turn off deprecation warnings when compiling class implementation in favor
+// of deprecating TexRefInputIterator instead.
```
brycelelbach, line 76: I don't understand this comment, why "in favor of deprecating TexRefInputIterator"? Don't we disable warnings here so that they don't trigger in our own code, just on external usage?

brycelelbach (approved):

cub/iterator/tex_ref_input_iterator.cuh
allisonvacanti: Right.

cub/iterator/tex_ref_input_iterator.cuh
allisonvacanti: The internal usages of deprecated CUDART APIs are always suppressed, and instead we deprecate the CUB utilities that call them. So instead of emitting deprecation warnings for `cudaBindTexture`, etc, we deprecate `cub::TexRefInputIterator` directly.

PR #325, opened 2021-06-18
Avoid warnings about unreachable code

dkolsen-pgi:
NVC++ warns about unreachable code in specializations of `BaseDigitExtractor::ProcessFloatMinusZero` where `FLOAT_KEY` is false.  Rearrange the code to avoid the warning, without changing the behavior at all.

### DIFF
--- a/cub/block/radix_rank_sort_operations.cuh
+++ b/cub/block/radix_rank_sort_operations.cuh
@@ -93,12 +93,14 @@ struct BaseDigitExtractor
 
     static __device__ __forceinline__ UnsignedBits ProcessFloatMinusZero(UnsignedBits key)
     {
-        if (!FLOAT_KEY) return key;
-        
-        UnsignedBits TWIDDLED_MINUS_ZERO_BITS =
-            TraitsT::TwiddleIn(UnsignedBits(1) << UnsignedBits(8 * sizeof(UnsignedBits) - 1));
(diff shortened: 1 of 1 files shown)


dkolsen-pgi:
This is not urgent.  Wait until after the 1.13 release is out the door before merging it.  But it is simple enough that I would like it merged before too long.

allisonvacanti:
It seems like the compiler shouldn't warn on this, since the condition is parametrized and either branch could be valid. Why not just discard the dead code silently?

LGTM otherwise. Since it's a harmless change I'll go ahead and test / merge this once gpuCI is working again, but this is surprising behavior.

dkolsen-pgi:
You are correct that the compiler shouldn't warn about this.  Before creating this PR, I filed a bug against EDG (our compiler front end) to get rid of the warning.  It will take a while for the compiler to get fixed, so I would like to change the CUB source to avoid the warning in the meantime.

allisonvacanti:
Gotcha -- sounds good to me.

Rebased to pull in recent CI fixes.

DVS CL: 30167030
gpuCI: NVIDIA/thrust#1482

allisonvacanti:
Rebased for more CI fixes...

DVS CL: 30211051
gpuCI: NVIDIA/thrust#1482

allisonvacanti (approved):
Tests pass, LGTM!

PR #326, opened 2021-06-18
Simplify usage of namespace macros, add thrust compatibility.

allisonvacanti:
This provides a workaround for downstream projects that encounter
a variety of issues from dynamically linking multiple libraries that
use Thrust and CUB.

See the new `cub/util_namespace.h` header for details.

Added several tests and checks to validate that this behavior is correct.

New tests:
  - test/test_namespace_wrapped.cu
  - test/cmake/check_namespace.cmake

This is change requires NVIDIA/thrust#1464.

### DIFF
--- a/cub/util_device.cuh
+++ b/cub/util_device.cuh
@@ -46,11 +46,7 @@
 #include <cassert>
 #endif
 
-/// Optional outer namespace(s)
-CUB_NS_PREFIX
-
-/// CUB namespace
-namespace cub {
+CUB_NAMESPACE_BEGIN
 
 
 /**
@@ -402,6 +398,7 @@ CUB_RUNTIME_FUNCTION inline cudaError_t PtxVersionUncached(int& ptx_version)
 __host__ inline cudaError_t PtxVersionUncached(int& ptx_version, int device)
 {
     SwitchDevice sd(device);
+    (void)sd;
     return PtxVersionUncached(ptx_version);
 }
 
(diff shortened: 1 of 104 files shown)


cub/util_device.cuh
```diff
@@ -402,6 +398,7 @@ CUB_RUNTIME_FUNCTION inline cudaError_t PtxVersionUncached(int& ptx_version)
 __host__ inline cudaError_t PtxVersionUncached(int& ptx_version, int device)
 {
     SwitchDevice sd(device);
+    (void)sd;
```
senior-zero, line 401: SwitchDevice has the non-trivial constructor and destructor. Its object [won't cause unused variable warnings](https://godbolt.org/z/br344eE1b) in this context. Have you encountered an issue here?

senior-zero (commented):
All of the header files contain the new namespace macro. Ones that don't contain it don't require it (macro + includes):
```bash
cub> find . -type f | wc -l
96
cub> find . -name "*cuh" -exec grep -l 'CUB_NAMESPACE_BEGIN' {} \; | wc -l
88
cub> find . -name "*cuh" -exec grep -l 'CUB_NAMESPACE_END' {} \; | wc -l
88
```
The same applies to the inclusion of `util_namespace.cuh` which can be found in `config.cuh`:
```bash
find . -name "*cuh" -exec grep -L 'config\.cuh' {} \; | wc -l
14
```

Headers that don't include `config.h` include `util_namespace.cuh` directly.

So it seems good to me. Thank you for this work! 

There is one minor question that can be skipped, though.

cub/util_device.cuh
allisonvacanti: I can't remember if I actually hit an error here, but either way I still prefer to explicitly mark these as used. Unused variable warnings are notorious for false positives, so adding these defensively helps avoid situations where a new compiler release will create issues for our users.

PR #331, opened 2021-06-29
Port adjacent difference into CUB

senior-zero:
The PR contains:

1. a port of `thrust::adjacent_difference` algorithm;
2. deprecation of `FlagHeads` and `FlagTails` methods from `BlockAdjacentDifference` structure;
3. fixed API for `BlockAdjacentDifference` along with documentation and tests.

Note that the PR is based on some of the features introduced in `MergeSort` porting.

### DIFF
--- a/cub/agent/agent_adjacent_difference.cuh
+++ b/cub/agent/agent_adjacent_difference.cuh
@@ -1,0 +1,254 @@
+/******************************************************************************
+ * Copyright (c) 2011-2021, NVIDIA CORPORATION.  All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the NVIDIA CORPORATION nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL NVIDIA CORPORATION BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ ******************************************************************************/
+
+#pragma once
+
+#include "../config.cuh"
+#include "../util_type.cuh"
+#include "../util_namespace.cuh"
+#include "../block/block_load.cuh"
+#include "../block/block_store.cuh"
+#include "../block/block_adjacent_difference.cuh"
+
+#include <thrust/system/cuda/detail/core/util.h>
+
+
+CUB_NAMESPACE_BEGIN
+
+
+template <
+  int                      _BLOCK_THREADS,
+  int                      _ITEMS_PER_THREAD = 1,
+  cub::BlockLoadAlgorithm  _LOAD_ALGORITHM   = cub::BLOCK_LOAD_DIRECT,
+  cub::CacheLoadModifier   _LOAD_MODIFIER    = cub::LOAD_LDG,
+  cub::BlockStoreAlgorithm _STORE_ALGORITHM  = cub::BLOCK_STORE_DIRECT>
+struct AgentAdjacentDifferencePolicy
+{
+  static constexpr int BLOCK_THREADS    = _BLOCK_THREADS;
+  static constexpr int ITEMS_PER_THREAD = _ITEMS_PER_THREAD;
+  static constexpr int ITEMS_PER_TILE   = BLOCK_THREADS * ITEMS_PER_THREAD;
+
+  static constexpr cub::BlockLoadAlgorithm LOAD_ALGORITHM   = _LOAD_ALGORITHM;
+  static constexpr cub::CacheLoadModifier LOAD_MODIFIER     = _LOAD_MODIFIER;
+  static constexpr cub::BlockStoreAlgorithm STORE_ALGORITHM = _STORE_ALGORITHM;
+};
+
+template <typename Policy,
+          typename InputIteratorT,
+          typename OutputIteratorT,
+          typename DifferenceOpT,
+          typename OffsetT,
+          typename InputT,
+          typename OutputT,
+          bool InPlace,
+          bool ReadLeft>
+struct AgentDifference
+{
+  using LoadIt = typename THRUST_NS_QUALIFIER::cuda_cub::core::LoadIterator<Policy, InputIteratorT>::type;
+
+  using BlockLoad = typename cub::BlockLoadType<Policy, LoadIt>::type;
+  using BlockStore = typename cub::BlockStoreType<Policy, OutputIteratorT, OutputT>::type;
+
+  using BlockAdjacentDifferenceT =
+    cub::BlockAdjacentDifference<InputT, Policy::BLOCK_THREADS>;
+
+  union _TempStorage
+  {
+    typename BlockLoad::TempStorage load;
+    typename BlockStore::TempStorage store;
+    typename BlockAdjacentDifferenceT::TempStorage adjacent_difference;
+  };
+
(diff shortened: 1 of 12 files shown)


test/test_block_adjacent_difference.cu
```diff
@@ -0,0 +1,521 @@
...
+                                      ItemsPerThread,
+                                      ThreadsInBlock,
+                                      read_left>(thrust::raw_pointer_cast(
+                                                   d_data.data()),
+                                                 num_items);
+
+  {
+    const std::size_t expected_count = num_items - 1;
+    const std::size_t actual_count =
+      thrust::count(d_data.begin(), d_data.end(), static_cast<DataType>(1));
```
allisonvacanti, line 263: I'm a little nervous about having so many tests producing a uniform output, both for the block and device tests. There are a lot of potential issues that these wouldn't catch (transposed outputs, invalid index calculations, etc). We should add some tests that produce a predictable but non-uniform result.

For instance, we could use an input array where `in[i] = (i * (i - 1)) / 2`. Taking the adjacent difference of that will always produce a sequence:

```
idx:       |  0  1  2  3  4  5  6 ...  N        |
input:     |  0  0  1  3  6 10 15 ...  N(N-1)/2 |
sub_left:  |  0  0  1  2  3  4  5 ...  N-1      |
sub_right: |  0 -1 -2 -3 -4 -5 -6 ...  N(N-1)/2 |
```

test/test_block_adjacent_difference.cu
```diff
@@ -0,0 +1,521 @@
...
+    d_values.resize(num_items);
+    TestLastTile<ItemsPerThread, ThreadsInBlock>(num_items, d_values);
+  }
+
+  d_values.resize(tile_size);
+  TestBaseCase<ItemsPerThread, ThreadsInBlock>(tile_size, d_values);
+}
+
+template <unsigned int ItemsPerThread, unsigned int ThreadsPerBlock>
+void Test()
```
allisonvacanti, line 491: Reminder to test in-place `BlockAdjacentDifference` after updating the implementation/docs to support this.

cub/agent/agent_adjacent_difference.cuh
```diff
@@ -0,0 +1,272 @@
...
+          typename OutputIteratorT,
+          typename FlagOpT,
+          typename OffsetT,
+          typename InputT,
+          bool InPlace,
+          bool ReadLeft>
+struct AgentDifference
+{
+  // XXX output type must be result of BinaryOp(input_type,input_type);
+  using OutputT = InputT;
```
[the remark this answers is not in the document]
allisonvacanti: (feel free to either add the `cub/detail/type_traits.cuh` header in this PR or just replicate the logic inline)

cub/device/dispatch/dispatch_adjacent_difference.cuh
```diff
@@ -0,0 +1,346 @@
...
+                                             OffsetT,
+                                             ValueT,
+                                             InPlace,
+                                             ReadLeft>;
+
+    cudaError error = cudaSuccess;
+
+    do
+    {
+      OffsetT tile_size = AdjacentDifferencePolicyT::ITEMS_PER_TILE;
```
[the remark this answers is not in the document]
allisonvacanti: The APIs that consume `tile_size` also appear to expect `int`.

allisonvacanti (changes requested):
This is a great first pass 👍 Let me know when you're ready for a re-review.

cub/device/device_adjacent_difference.cuh
```diff
@@ -0,0 +1,717 @@
...
+   * \tparam OutputIteratorT is a model of <a href="https://en.cppreference.com/w/cpp/iterator/output_iterator">Output Iterator</a>.
+   * \tparam DifferenceOpT's \c result_type is convertible to a type in \p OutputIteratorT's set of \c value_types.
+   * \tparam OffsetT is an integer type for global offsets.
+   */
+  template <typename InputIteratorT,
+            typename OutputIteratorT,
+            typename DifferenceOpT,
+            typename OffsetT>
+  static CUB_RUNTIME_FUNCTION cudaError_t
+  SubtractLeftCopy(InputIteratorT d_input,          ///< [in] Pointer to the input sequence
```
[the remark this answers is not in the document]
senior-zero: > I still wish we could find a way to make this take temporary storage, just for API consistency and future-proofing.
> 
> Unfortunately `cudaMalloc(&ptr, 0)` sets `ptr -> nullptr` so we wouldn't be able to detect which pass we're in. The only option I see is to force the user to make a small allocation that we'd never use, but that's horrible.
> 
> I suppose we'll just have to live with the inconsistency :(

There's a case like that in CUB already (`DispatchRadixSort`):
```cuda
if (d_temp_storage == NULL)
{
    temp_storage_bytes = 1;
    break;
}
```
We could use this approach for consistency.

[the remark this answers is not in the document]
allisonvacanti: Interesting, I hadn't seen that before. Good catch!

Let's do that since it's already existing practice. It'll be useful if the implementation ever changes to require temp storage.

cub/block/block_adjacent_difference.cuh
```diff
@@ -239,11 +307,357 @@ public:
...
+     *         thread_data,
+     *         CustomDifference(),
+     *         valid_items);
+     *
+     * \endcode
+     * \par
+     * Suppose the set of input \p thread_data across the block of threads is
+     * <tt>{ ...3], [4,2,1,1], [1,1,1,1], [2,3,3,3], [3,4,1,4] }</tt>.
+     * and that \p valid_items is \p 507. The corresponding output \p result in those threads will be
+     * <tt>{ ..., [-1,2,1,0], [0,0,0,-1], [-1,0,3,3], [3,4,1,4] }</tt>.
```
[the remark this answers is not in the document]
senior-zero: Thank you so much for being attentive! As I can see, there are `128` threads and `4` items per thread in the example. That's `512` items per thread block, `517` of which are valid (have neighbour value to the right). So, the last thread doesn't have valid items at all. The pre-last thread has 3 valid items. As the documentation states, the input value is copied without modifications if the neighbour value is out of bounds. As I can see, the example conforms this description.

```
[ -1, 0, /* last valid item, has out of bounds neighbour to the right => */ input[i], input[i + 1]], 
[ input[i + 2], input[i + 3], input[i + 4], input[i + 5] ]
```

Therefore, `(128 * 4 - 507 /*=5*/) + 1` items copied. Is there anything I'm missing?

cub/agent/agent_adjacent_difference.cuh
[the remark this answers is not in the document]
senior-zero: I'll copy your `cub/detail/type_traits.cuh` if you don't mind.

cub/block/block_adjacent_difference.cuh
```diff
@@ -200,50 +294,684 @@ public:
...
+     *         CustomDifference(),
+     *         valid_items);
+     *
+     * @endcode
+     * @par
+     * Suppose the set of input `thread_data` across the block of threads is
+     * `{ ...3], [4,2,1,1], [1,1,1,1], [2,3,3,3], [3,4,1,4] }`.
+     * and that `valid_items` is `507`. The corresponding output `result` in
+     * those threads will be
+     * `{ ..., [-1,2,1,0], [0,0,0,-1], [-1,0,3,3], [3,4,1,4] }`.
```
allisonvacanti, line 892: (continuing from https://github.com/NVIDIA/cub/pull/331#discussion_r674320038)

Gotcha, so the last 5 values are unchanged random values that are ignored by the algorithm.

There's another convention that uses `-` to stand in for such values, e.g. https://github.com/NVIDIA/cub/blob/d6ba3cf0d38e1ab3f079b1f301551e7d4d2ec8e4/cub/device/device_histogram.cuh#L181-L182

I think it's a little easier to parse with the non-numeric characters, since it's immediately clear that they _can't_ be involved in the calculation. If you agree, let's change this to

```
* `{ ...3], [4,2,1,1], [1,1,1,1], [2,3,3,-], [-,-,-,-] }`.
* `{ ..., [-1,2,1,0], [0,0,0,-1], [-1,0,3,-], [-,-,-,-] }`.
```

cub/block/block_adjacent_difference.cuh
```diff
@@ -39,27 +41,105 @@
...
+ *   __device__ DataType operator()(DataType &lhs, DataType &rhs)
+ *   {
+ *     return lhs - rhs;
+ *   }
+ * };
+ *
+ * __global__ void ExampleKernel(...)
+ * {
+ *     // Specialize BlockAdjacentDifference for a 1D block of
+ *     // 128 threads on type int
```
allisonvacanti, line 99: s/on/of/

(this is repeated in a few other docstrings)

allisonvacanti (approved):
LGTM -- just some comments on documentation, otherwise this is ready to start testing 👍

cub/block/block_adjacent_difference.cuh
senior-zero: > Gotcha, so the last 5 values are unchanged random values that are ignored by the algorithm.

These values aren't ignored by the algorithm:
> As the documentation states, the input value is copied without modifications if the neighbour value is out of bounds.

If `input` and `output` parameters point to different memory locations, these values will be copied without modification. I'm afraid that non-numeric characters would mean that the output values are unchanged at all, for example:
```
input { 1, 2, 3 };
output { 4, 5, 6 };
valid_items = 0;

adjacent_difference(input, output, valid_items);

// output should be { 1, 2, 3 } and not { 4, 5, 6 }
```

The same with `std::adjacent_difference` [documentation](https://en.cppreference.com/w/cpp/algorithm/adjacent_difference). The output for:
```
std::vector v {2, 4, 6, 8, 10, 12, 14, 16, 18, 20};
```

is 
```
2 2 2 2 2 2 2 2 2 2
```

rather than:

```
- 2 2 2 2 2 2 2 2 2
```

Do you think it'll be clear from the `-` notation that the values are copied?

cub/block/block_adjacent_difference.cuh
senior-zero: There are quite a few places with this typo:
```
rg "threads on type" | wc -l
33
```

I'll fix this in the `block_scan`, `block_discontinuity` and `block_reduce` documentation as well.

cub/block/block_adjacent_difference.cuh
allisonvacanti: That's a good point -- I'm fine leaving this as-is.

senior-zero:
gpuCI: NVIDIA/thrust/pull/1577
DVS: 30766806

PR #338, opened 2021-07-01
Make cub::InitValue (test_util.h) host-only.

allisonvacanti:
This is never called from the device, and many of the implementations
use a host-only random number generator, which is silently ignored
in the device passes.

This is needed to fix gpuCI builds for Thrust.

### DIFF
--- a/test/test_util.h
+++ b/test/test_util.h
@@ -542,11 +542,10 @@ enum GenMode
  * Initialize value
  */
 template <typename T>
-__host__ __device__ __forceinline__ void InitValue(GenMode gen_mode, T &value, int index = 0)
+void InitValue(GenMode gen_mode, T &value, int index = 0)
 {
     switch (gen_mode)
     {
-#if (CUB_PTX_ARCH == 0)
     case RANDOM:
         RandomBits(value);
         break;
@@ -581,7 +580,6 @@ __host__ __device__ __forceinline__ void InitValue(GenMode gen_mode, T &value, i
         }
         break;
     }
-#endif
      case UNIFORM:
         value = 2;
         break;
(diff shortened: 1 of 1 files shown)


test/test_util.h
```diff
@@ -542,11 +542,10 @@ enum GenMode
  * Initialize value
  */
 template <typename T>
-__host__ __device__ __forceinline__ void InitValue(GenMode gen_mode, T &value, int index = 0)
```
senior-zero, line 545: Isn't it used in the host/device method of `struct TransformOp` in test/test_iterator.cu?

senior-zero: Regarding test/test_iterator.cu. It seems to be using deprecated methods for TestTex*.

senior-zero (changes requested):
It seems that we have to fix `test/test_iterator.cu` first. Apart from that, everything looks good. Thank you for fixing the build!

test/test_util.h
allisonvacanti: > https://github.com/NVIDIA/cub/pull/338

Yeah -_- CLion missed that usage. I'll have to approach this differently. I'll probably just #ifdef out the random usage that the compiler was complaining about and add some warnings.

> Regarding test/test_iterator.cu. It seems to be using deprecated methods for TestTex*.

This is fine. We build tests with `CUB_IGNORE_DEPRECATED_API` so that we can still test deprecated stuff.

test/test_util.h
```diff
@@ -622,7 +618,7 @@ __host__ __device__ __forceinline__ void InitValue(GenMode gen_mode, bool &value
 /**
  * cub::NullType test initialization
  */
-__host__ __device__ __forceinline__ void InitValue(GenMode /* gen_mode */,
+void InitValue(GenMode /* gen_mode */,
 						   cub::NullType &/* value */,
```
senior-zero, line 622: It might be worth reformating this code fragment.

allisonvacanti:
Closing -- I'll have another PR up soon.

PR #339, opened 2021-07-01
gpuCI fixes

allisonvacanti:


allisonvacanti:
@senior-zero This is ready for review when we come back next week.

There are still a couple failing builds on gpuCI, but they can be fixed in a follow up. This at least restores gpuCI to where it was before the upgrade and unblocks other PRs.

gpuCI results in NVIDIA/thrust#1474:
- nvc++ build has always failed and has configuration issues, it will take significant effort to fix.
- gcc-10 build is ICE'ing. This is a new build for our CI and will take more investigation to find a WAR.

Started DVS testing in CL 30149184.

test/test_util.h
```diff
@@ -581,6 +584,14 @@ __host__ __device__ __forceinline__ void InitValue(GenMode gen_mode, T &value, i
         }
         break;
     }
+#else
+     case RANDOM:
+     case RANDOM_BIT:
+     case RANDOM_MINUS_PLUS_ZERO:
+         _CubLog("%s\n",
+                 "cub::InitValue cannot generate random numbers on device.");
+         cub::ThreadTrap();
```
senior-zero, line 593: I understand that we have to fix the build as soon as possible, so the following suggestion could be extracted into a separate issue. We could split this code into device/host supported parts instead of failing at runtime. Here is the prototype that I've been thinking about:

```cuda
enum GenModeTrivialParam
{
    UNIFORM = 0,                 // Assign to '2', regardless of integer seed
    INTEGER_SEED = 1             // Assign to integer seed
};

struct GenModeTrivial
{
    GenModeTrivialParam param;

    GenModeTrivial(GenModeTrivialParam param)
        : param(param)
    {
    }

    operator int()
    {
        return param;
    }
};

enum GenModeRandomParam
{
    RANDOM = 2,                  // Assign to random, regardless of integer seed
    RANDOM_BIT = 3,              // Assign to randomly chosen 0 or 1, regardless of integer seed
    RANDOM_MINUS_PLUS_ZERO = 4,  // Assign to random, with some values being -0.0 or +0.0 patterns
};

struct GenModeRandom
{
    GenModeRandomParam param;

    GenModeRandom(GenModeRandomParam param)
        : param(param)
    {
    }

    operator int()
    {
        return param;
    }
};

/**
 * Initialize value
 */
template <typename T>
void InitValue(GenModeRandom gen_mode, T &value, int index = 0)
{
    switch (gen_mode.param)
    {
      // RandomBits is host-only.
    case RANDOM:
        RandomBits(value);
        break;
    case RANDOM_BIT:
    {
        char c;
        RandomBits(c, 0, 0, 1);
        value = (c > 0) ? (T) 1 : (T) -1;
        break;
    }
    case RANDOM_MINUS_PLUS_ZERO:
    {
        // Replace roughly 1/128 of values with -0.0 or +0.0, and generate the rest randomly
        typedef typename cub::Traits<T>::UnsignedBits UnsignedBits;
        char c;
        RandomBits(c);
        if (c == 0)
        {
            // Replace 1/256 of values with +0.0 bit pattern
            value = SafeBitCast<T>(UnsignedBits(0));
        }
        else if (c == 1)
        {
            // Replace 1/256 of values with -0.0 bit pattern
            value = SafeBitCast<T>(UnsignedBits(UnsignedBits(1) <<
                                                (sizeof(UnsignedBits) * 8) - 1));
        }
        else
        {
            // 127/128 of values are random
            RandomBits(value);
        }
        break;
    }
    default:
         value = (T) index;
        break;
    }
}

template <typename T>
__device__ __host__ void InitValue(GenModeTrivial gen_mode, T &value, int index = 0)
{
    switch (gen_mode.param)
    {
        // RandomBits is host-only.
        case UNIFORM:
            value = 2;
            break;
        case INTEGER_SEED:
        default:
            value = (T) index;
            break;
    }
}
```

There is little to be changed in the tests. I've tested this modification on the `test/test_block_histogram.cu`, here are the changes I've made: 

```cuda
template <
    int             BINS,
    typename        SampleT,
    typename        GenMode>  // <-------------------------------------
void Initialize(

///.....

template <
    typename                    SampleT,
    int                         BINS,
    int                         BLOCK_THREADS,
    int                         ITEMS_PER_THREAD,
    BlockHistogramAlgorithm     ALGORITHM,
    typename                    GenMode> // <-------------------------------------
void Test(
```

In most cases, gen_mode is selected in place of InitValue call, so there won't be any changes at all:

```cuda
InitValue(INTEGER_SEED, h_in[i], i);
```

As an alternative, we could leave only host versions of `InitValue` (as in your previous PR) and copy an INTEGER_SEED version for `template <typename T> struct TransformOp`. It seems that it's the only place where InitValue is called from the device side. 

In both cases, we'll fail at the compilation time instead of runtime. We'll also have simpler code.

senior-zero (changes requested):
All of the comments are optional, so if you think that we don't have time to fix them at the moment, it's fine to merge this PR :smiley:

cmake/CubBuildCompilerTargets.cmake
```diff
@@ -110,10 +110,15 @@ function(cub_build_compiler_targets)
...
-    $<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CUDA_COMPILER_ID:NVIDIA>>:-Xcudafe=--promote_warnings>
+    $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:-Xcudafe=--promote_warnings>
     # Don't complain about deprecated GPU targets.
-    $<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CUDA_COMPILER_ID:NVIDIA>>:-Wno-deprecated-gpu-targets>
+    $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:-Wno-deprecated-gpu-targets>
+    # Suppress deprecation warnings in nvcc < 11.5.
+    # TexRefInputIterator uses deprecated CUDART APIs, see NVIDIA/cub#191.
+    # After 11.5, we will suppress these in-code via pragma, but for older nvcc
+    # we have to use the big hammer:
+    $<$<AND:$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>,$<VERSION_LESS:$<CUDA_COMPILER_VERSION>,11.5>>:-Wno-deprecated-declarations>
```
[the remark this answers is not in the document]
allisonvacanti: This is a good point. There are some definite advantages to just removing the deprecated class, but it's still a bit early IMO.

While CUB's version has increased since we deprecated this, we haven't shipped an updated version of CUB in the CTK yet. Depending on how future CUB/CTK release schedules line up, there might not be enough time for users to react to the deprecation if we remove the texture ref stuff now.

Let's give it a couple more releases before removing this entirely.

test/test_util.h
```diff
@@ -631,6 +651,7 @@ __host__ __device__ __forceinline__ void InitValue(GenMode /* gen_mode */,
 /**
  * cub::KeyValuePair<OffsetT, ValueT>test initialization
  */
+#pragma nv_exec_check_disable
```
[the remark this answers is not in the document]
allisonvacanti: I agree, this is hacky and not-great. It definitely makes sense to refactor this as we improve CUB's tests. Maybe something like thrust_benchmark's `range_generator` could be used instead?

But since it's just a testing facility with no user exposure, this is very low priority. We can look into this while doing test clean-ups, but for now let's just get things building again to unblock gpuCI 👍

test/test_util.h
allisonvacanti: We could also switch to an RNG implementation that works on the device. This will have to wait for a follow-up fix though.

dkolsen-pgi, line 593: This breaks NVC++.  A bunch of the NVHPC tests have been failing since this was checked in.

This change uses the NVCC split compilation model.  It doesn't work with NVC++'s unified compilation model.  NVC++ only sees the device version of the random options, and will use that for both host and device.  Not only will it not run correctly, it won't even compile, because it contains a call to the device-only function `cub::ThreadTrap` which causes failures during host code generation.

This change needs to use a host/device mechanism that works with both NVCC and NVC++.  Either `<nv/target>` or the older (and much more cumbersome) `if (CUB_IS_HOST_CODE)`.

allisonvacanti: d'oh, I should have thought of this. I'll rewrite this to use the `CUB_IS_HOST_CODE` pattern for now since the `<nv/target>` refactor is still a WIP.

Issue #344, opened 2021-07-08
BlockReduce incorrect result / ignoring certain thread values


MatthiasKohl:
I have an example where I noticed cub is giving wrong results.
On further investigation, it looks like the `BlockReduce::Sum` function from cub is "ignoring" values from certain threads, but it's a bit erratic.
Here is the gist, with an example with 64 threads / block and 33 threads holding non-zero values:
https://gist.github.com/MatthiasKohl/f84cfba617ec6edf6a21e9bbc69f57fe

I'm using Ubuntu 16.04 (64-bit) and CUDA 11.3, see the exact version and compile command in the gist.
It feels like a compiler issue, but I don't have time to investigate any deeper.

Specifying how many threads participate in the reduction does not change the results, for me.

EDIT: forgot to mention that the tests I did were on V100 (PCIE 16GB)

senior-zero:
Hello, @MatthiasKohl! 

There are two issues in your gist. First of all, it should be

```cuda
auto &red_storage = *reinterpret_cast<typename BlockReduce::TempStorage*>(smem);
```

rather than:
```cuda
auto red_storage = *reinterpret_cast<typename BlockReduce::TempStorage*>(smem);
```

You've created a thread-local copy of TempStorage, so threads of the same block couldn't communicate with each other using it. Using the reference fixes your case. 

Apart from that, you [can't estimate](https://forums.developer.nvidia.com/t/how-can-one-determine-the-temp-storage-size-for-cub-block-primitives-from-the-host/112839) the temporary storage size from the host side, because CUB uses `CUB_PTX_ARCH` macro. As a workaround you could manually specify the architecture:

```cuda
auto smem_size = sizeof(typename cub::BlockReduce<float, TPB, cub::BLOCK_REDUCE_WARP_REDUCTIONS, 1, 1, 700>::TempStorage);

```

With these fixes, your gist works fine, so I'm closing the issue. Please, feel free to reopen it if there is anything I've overlooked.

MatthiasKohl:
Thanks for looking into this!
May I suggest adding an overload of the constructor for `BlockReduce` which works somehow nicely with dynamic shared memory?
It's a common use-case in order to re-use shared memory, but to me, it seems like it's not really made for this.

gonzalobg:
> May I suggest adding an overload of the constructor for BlockReduce which works somehow nicely with dynamic shared memory?

[...]

PR #348, opened 2021-07-16
Docs update: how to use dynamic shared mem with block reduce

MatthiasKohl:
As mentioned in https://github.com/NVIDIA/cub/issues/344, I created this PR to add a simple example of how to correctly use dynamic shared memory with `BlockReduce`.

I hope I followed the PR process correctly, I wasn't able to re-build the docs because the full build (for thrust) took longer than my allocated time on the machine when I was working on this, and I'm not familiar enough with the build process s.t. I can simply only build the docs for cub.
I'll amend the pull request, of course, if needed.

Plus, do you want me to add a similar example to classes like `BlockScan` ? In this case, it may be better to add an actual example file and link to it from the various inline docs.

### DIFF
--- a/cub/block/block_discontinuity.cuh
+++ b/cub/block/block_discontinuity.cuh
@@ -94,6 +94,13 @@ CUB_NAMESPACE_BEGIN
  * \par Performance Considerations
  * - Incurs zero bank conflicts for most types
  *
+ * \par Re-using dynamically allocating shared memory
+ * The following example under the examples/block folder illustrates usage of
+ * dynamically shared memory with BlockReduce and how to re-purpose
+ * the same memory region:
+ * <a href="../../examples/block/example_block_reduce_dyn_smem.cu">example_block_reduce_dyn_smem.cu</a>
+ *
+ * This example can be easily adapted to the storage required by BlockDiscontinuity.
  */
 template <
     typename    T,
--- a/cub/block/block_exchange.cuh
+++ b/cub/block/block_exchange.cuh
@@ -98,6 +98,13 @@ CUB_NAMESPACE_BEGIN
  * \par Performance Considerations
  * - Proper device-specific padding ensures zero bank conflicts for most types.
  *
+ * \par Re-using dynamically allocating shared memory
+ * The following example under the examples/block folder illustrates usage of
+ * dynamically shared memory with BlockReduce and how to re-purpose
+ * the same memory region:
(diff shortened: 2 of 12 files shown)


allisonvacanti:
> I wasn't able to re-build the docs

No problem, our docs framework is a work-in-progress. This should improve soon.

> it may be better to add an actual example file and link to it from the various inline docs.

I agree, this would be better to have a complete example under the `examples/` directory. We'll build it as part of our CI to make sure it keeps working this way, too. Can you add this?

MatthiasKohl:
> I agree, this would be better to have a complete example under the `examples/` directory. We'll build it as part of our CI to make sure it keeps working this way, too. Can you add this?

Sure, I'll add it as soon as I get some time for it (probably tomorrow).

MatthiasKohl:
Sorry for the delay, but I finally got to doing this and added the full example.
I can see that the example compiles (and tests pass), but I'm not sure what the best way to integrate it into the docs is, yet.
Especially since I wasn't able to compile the docs.
Would it be possible for you to check that? It may also make sense to link to this example from other block-wide operations since they'll work basically the same w.r.t. shared memory

allisonvacanti:
Unfortunately, our docs build is broken at the moment, so we can't really test this out just yet. NVIDIA/thrust#1475 is working on restoring it.

It might be best to just mention the name of the example without a link. Doxygen provides the `\ref` command to create internal links, but I don't see how to get that to link to a complete source file, it appears to only work with documentation blocks.

> It may also make sense to link to this example from other block-wide operations since they'll work basically the same w.r.t. shared memory

Agreed -- could you update this PR to do so?

MatthiasKohl:
Yes, I slightly reformulated the docs such that it should be clear where to find the example in case the <a/> tags don't work.
Although I believe they should work with doxygen and I've seen other projects using them (they are even supposed to be automatically substituted with references if you build the docs to LaTeX, but I'll believe it when I see it...).

And I added this example along with the short description to all block-wide operations except for `BlockAdjacentDifference` which doesn't seem to have any docs for now, and `BlockShuffle` which has very minimal docs.

allisonvacanti:
@senior-zero is currently working on `BlockAdjacentDifference` -- George, can you include similar wording as the new docs here?

I should be able to get this merged in the next few days.

allisonvacanti:
Made some minor modifications. Started tests in NVIDIA/thrust#1504.

allisonvacanti (approved):

PR #349, opened 2021-07-17
Remove extra % character and some unused variables

senior-zero:
I've noticed confusing SM number queries in algorithms that shouldn't use it. So I decided to combine this refactoring with the removal of `%Device` comments that we've discussed. This PR has the lowest possible priority.

### DIFF
--- a/cub/agent/single_pass_scan_operators.cuh
+++ b/cub/agent/single_pass_scan_operators.cuh
@@ -177,7 +177,7 @@ struct ScanTileState<T, true>
     __host__ __device__ __forceinline__
     cudaError_t Init(
         int     /*num_tiles*/,                      ///< [in] Number of tiles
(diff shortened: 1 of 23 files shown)


allisonvacanti (approved):
LGTM -- go ahead and start testing on this one.

(I removed the `only: docs` tag, since this does have code changes. That tag is for pure doc changes that don't require testing.)

allisonvacanti:
I just came across this in the [Doxygen FAQ](https://www.doxygen.nl/manual/faq.html):

> Q: Doxygen automatically generates a link to the class MyClass somewhere in the running text. How do I prevent that at a certain place?
A: Put a % in front of the class name. Like this: %MyClass. Doxygen will then remove the % and keep the word unlinked.

I don't see how this would apply here, since we don't have anything just named "Device". Maybe there used to be?

Either way, we should keep an eye on this once we have the new docs site up.

allisonvacanti:
@senior-zero Since this is so close to being ready, let's get it into 1.15. Can you check on the DVS results for this and merge if all is well?

PR #350, opened 2021-07-27
Notify users that CUB_NS_QUALIFIER must be set when using other NS macros

allisonvacanti:


allisonvacanti:
DVS CL: 30228013
gpuCI: NVIDIA/thrust#1495

senior-zero (approved):
That's an important change!

PR #351, opened 2021-07-28
Update histogram documentation.

allisonvacanti:
- Remove extra values from `d_histogram` result.
- Update input samples to avoid fp values that are the same as
  a computed level. These introduce ambiguity into the result due
  to fp rounding/math.
  
  cc: @cliffburdick

### DIFF
--- a/cub/device/device_histogram.cuh
+++ b/cub/device/device_histogram.cuh
@@ -82,8 +82,8 @@ struct DeviceHistogram
      * // Declare, allocate, and initialize device-accessible pointers for input samples and
      * // output histogram
      * int      num_samples;    // e.g., 10
-     * float*   d_samples;      // e.g., [2.2, 6.0, 7.1, 2.9, 3.5, 0.3, 2.9, 2.0, 6.1, 999.5]
-     * int*     d_histogram;    // e.g., [ -, -, -, -, -, -, -, -]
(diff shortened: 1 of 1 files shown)


allisonvacanti:
@senior-zero This needs a review when you get a chance.

cub/device/device_histogram.cuh
```diff
@@ -178,9 +178,9 @@ struct DeviceHistogram
      * int      num_row_samples;    // e.g., 5
      * int      num_rows;           // e.g., 2;
      * size_t   row_stride_bytes;   // e.g., 7 * sizeof(float)
-     * float*   d_samples;          // e.g., [2.2, 6.0, 7.1, 2.9, 3.5,   -, -,
-     *                              //        0.3, 2.9, 2.0, 6.1, 999.5, -, -]
-     * int*     d_histogram;        // e.g., [ -, -, -, -, -, -, -, -]
```
senior-zero, line 183: It seems that the same fix is required for two overloads of `HistogramRange`.

senior-zero (approved):
That's definitely more unambiguous now, thanks! I think we should fix the remaining documentation and merge this PR.

cub/device/device_histogram.cuh
allisonvacanti: Oops, I thought I'd checked the other examples. I'll update this before merging.

PR #354, opened 2021-08-05
Adds BlockRunLengthDecode algorithm and tests

elstehle:
# Algorithm Overview

The `BlockRunLengthDecode` class supports decoding a run-length encoded array of `unique_items`. That is, given the two arrays `unique_items[N]` and `run_lengths[N]`,  `unique_items[i]` is repeated `run_lengths[i]` many times in the output array `decoded_items[M]`. *Note: runs of length `0` are supported and will not appear in the output.*
*The application of `BlockRunLengthDecode` goes beyond just the decompression use case. Two other use cases are load-balancing and generating the `keys` array of `*_by_key` algorithm variants. This is a preliminary building block for the BatchMemcpy https://github.com/NVIDIA/cub/issues/297.*

**Example**:
| unique_items    | 0 | 2 | 9 | 5 | 8 |   |   |   |
|-----------------|---|---|---|---|---|---|---|---|
| run_lengths     | 1 | 2 | 1 | 3 | 1 |   |   |   |
|                 |   |   |   |   |   |   |   |   |
| decoded_items   | 0 | 2 | 2 | 9 | 5 | 5 | 5 | 8 |
| decoded_size    | 8 |   |   |   |   |   |   |   |

**Specialisation that will also output relative offsets**:
*The relative offset indicates the offset **within** its run for each decoded item*
| unique_items    | 0 | 2 | 9 | 5 | 8 |   |   |   |
|-----------------|---|---|---|---|---|---|---|---|
| run_lengths     | 1 | 2 | 1 | 3 | 1 |   |   |   |
|                 |   |   |   |   |   |   |   |   |
| decoded_items   | 0 | 2 | 2 | 9 | 5 | 5 | 5 | 8 |
| relative_offsets   | 0 | 0 | 1 | 0 | 0 | 1 | 2 | 0 |
| decoded_size    | 8 |   |   |   |   |   |   |   |

In the fashion of CUB's block-level algorithms, the items being contributed by each thread are expected in a blocked arrangement. The items being returned are also returned in a blocked arrangement. 

Due to the nature of the run-length decoding algorithm ("decompression"), the output size of the algorithm invocation is unbounded. To address this, `BlockRunLengthDecode` allows retrieving a "window" from the run-length decoded output. The window's offset can be specified by the user and `BLOCK_THREADS * DECODED_ITEMS_PER_THREAD` (i.e., referred to as `window_size`) decoded items from the specified window will be returned. Memory requirements are always just `O(window_size)`.

```cpp
// The data type of the items being run-length decoded (i.e., the output data type)
using UniqueItemT = uint16_t;
constexpr uint32_t BLOCK_DIM_X = 128; // the number of threads per block
constexpr uint32_t RUNS_PER_THREAD = 2; // the number of runs contributed by each thread
constexpr uint32_t DECODED_ITEMS_PER_THREAD = 4; // the number of decoded items returned to each thread
using BlockRunLengthDecodeT = cub::BlockRunLengthDecode<UniqueItemT, BLOCK_DIM_X, RUNS_PER_THREAD, DECODED_ITEMS_PER_THREAD>;

// temporary shared memory required by the block-level run-length decode for the its lifetime
__shared__ typename BlockRunLengthDecodeT::TempStorage tmp_storage;

// Instantiate the BlockRunLengthDecode and give it the memory it can work with
BlockRunLengthDecodeT run_length_decode(tmp_storage);

// Initialize the BlockRunLengthDecode with the runs that we want to run-length decode
UniqueItemT unique_items[RUNS_PER_THREAD];  // assume this holds values
RunLengthT run_lengths[RUNS_PER_THREAD];    // assume this holds run lengths
uint32_t decoded_size = 0U;
run_length_decode.Init(unique_items, run_lengths, decoded_size);

// Run-length decode ("decompress") the runs into a window buffer of limited size. This is repeated until all runs
// have been decoded.
uint32_t decoded_window_offset = 0U;
while (decoded_window_offset < decoded_size)
{
  RunLengthT relative_offsets[DECODED_ITEMS_PER_THREAD];
  UniqueItemT decoded_items[DECODED_ITEMS_PER_THREAD];

  // The number of decoded items that are valid within this window (aka pass) of run-length decoding
  uint32_t num_valid_items = decoded_size - decoded_window_offset;
  run_length_decode.RunLengthDecode(decoded_items, relative_offsets, decoded_window_offset);
  BlockStoreDecodedItemT(temp_storage.store_decoded_runs_storage)
    .Store(d_block_decoded_out + decoded_window_offset, decoded_items, num_valid_items);
}
```

## Use Cases
- **Load-balancing batched problems**: While individual problems from a batch can be processed independently, simply assigning one thread/warp to each problem has two shortcomings: (1) load imbalance: some problems may take way longer than others, leaving some threads fiddling fingers while others are hard at work, (2) limited parallelism: your batch size may be much smaller than the parallelism you could exploit, (3) non-coalesced memory accesses, as individual problems may reside far apart in memory. `RunLengthDecode` allows to evenly sub-divide these problems into equi-sized sub-problems.  
- **Generating keys for `*_by_keys` algorithm variants**
- **Compression / Decompression**

# Implementation Details

**Overview of processing stages:**
| unique_items    | 0 | 2 | 9 | 5 | 8 |   |   |   |
|------------------|---|---|---|---|---|---|---|---|
| run_lengths      | 1 | 2 | 1 | 3 | 1 |   |   |   |
| *[1] prefix sum* |   |   |   |   |   |   |   |   |
| run_offsets      | 0 | 1 | 3 | 4 | 7 |   |   |   |
| aggregate        | 8 |   |   |   |   |   |   |   |
|                  |   |   |   |   |   |   |   |   |
| *[2] init buffer* |   |   |   |   |   |   |   |   |
| indexes          | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 |
| decode_buffer.uniques | - | - | - | - | - | - | - | - |
| *[3] scatter * |   |   |   |   |   |   |   |   |
| decode_buffer.uniques | 0 | 2 | - | 9 | 5 | - | - | 8 |
| *[4] incl. scan* |   |   |   |   |   |   |   |   |
| decoded_items    | 0 | 2 | 2 | 9 | 5 | 5 | 5 | 8 |

1. An exclusive prefix-sum computes the index of the beginning of each run in the decoded output
2. We initialize the `decode_buffer` with `continuation-of-run` items. This is just to differentiate between items that have already been resolved and items that yet need to be "filled in".
3. Scatter the beginning of each run into the appropriate position in the decode buffer, which was computed in [1], if that run's length is not `0`.
4. Perform an inclusive prefix scan that uses the following `bin_op` (where `-` from the table above means `ìs_unresolved`)
```
__device__ __forceinline__ T operator()(const T &lhs, const T &rhs)
{
    // If the rhs is still unresolved, propagate the lhs (which may or may not be resolved _at this point_, but
    // eventually _some_ lhs will provide the answer)
    return rhs.is_unresolved ? lhs : rhs;
}
```
A question that arises is how to differentiate between an `unresolved` (i.e., `- ` and an already decoded (or "resolved") item. If there was a value representable by the data type of `unique_items` that will never appear in the user-provided input, we could simply use that to represent unresolved items. The other alternative is to make `decoded_items` temporarily a pair of {`unique_value`, `is_resolved`}. I had started with the former and later added the latter. The former is 10-20% faster but not so nice with regards to the interface. So currently, we only provide an interface for the latter.

If the user also wants to retrieve `relative_offsets`, then the pair of {`unique_value`, `is_resolved`} is becoming {`unique_value`, `relative_offset`} and the scan operator is becoming slightly more involved. 

So, essentially there's three specialisations (or instances) of the `BlockRunLengthDecode` which have a lot of overlap implementation-wise. They mostly vary in the amount of TempStorage they require and the `RunLengthDecode` member function signature: 
**`UNUSED` (regular run-length decoding, but the user has to tell us which value we can use to represent yet-unresolved items):**
```
RunLengthDecode(UniqueItemT (&decoded_items)[DECODED_ITEMS_PER_THREAD], UniqueItemT unused_item_that_can_be_used_to_indicate_unresolved, int32_t from_decoded_offset = 0)
```

**`NORMAL` (regular run-length decoding):**
```
RunLengthDecode(UniqueItemT (&decoded_items)[DECODED_ITEMS_PER_THREAD], int32_t from_decoded_offset = 0)
```

**`OFFSETS` (...):**
```
template<typename UserRelativeOffsetT>
RunLengthDecode(UniqueItemT (&decoded_items)[DECODED_ITEMS_PER_THREAD], UserRelativeOffsetT (&item_offsets)[DECODED_ITEMS_PER_THREAD], int32_t from_decoded_offset = 0)
```

Currently only **`OFFSETS` (...):** and **`NORMAL` (regular run-length decoding):** are supported. And which implementation the user wants is decided by passing a template parameter to `BlockRunLengthDecode`. 

**Question:**
I'm inclined to have three different super classes (one for each of above specialisations) with a CRTP base class. So far CUB has refrained from having *any* inheritance. But here, the specialisations not only differ in the implementation but also the interface they expose. So, I think different classes would be the cleanest way to express that?

Overall I tried to match the CUB style wherever possible. I just deviated and had decided to go with fixed-width types. 

## Performance 
These are some numbers from a `V100` when decoding `uint32_t` as the `unique_items`. 
```
Using device 0: Tesla V100-SXM2-32GB (PTX version 700, SM700, 80 SMs, 32199 free / 32510 total MB physmem, 898.048 GB/s @ 877000 kHz mem clock, ECC on)
MODE: offsets, RUNS_PER_THREAD: 1, DECODED_ITEMS_PER_THREAD: 1, THREADS_PER_BLOCK: 64, decoded size (bytes): 198000000, relative offsets (bytes): 198000000, time_size (ms): 0.043072, time_decode (ms): 0.804256, achieved decode BW (GB/s): 492.381
MODE: normal,  RUNS_PER_THREAD: 1, DECODED_ITEMS_PER_THREAD: 1, THREADS_PER_BLOCK: 64, decoded size (bytes): 198000000, relative offsets (bytes): 0, time_size (ms): 0.060064, time_decode (ms): 0.671776, achieved decode BW (GB/s): 294.741
MODE: offsets, RUNS_PER_THREAD: 1, DECODED_ITEMS_PER_THREAD: 3, THREADS_PER_BLOCK: 192, decoded size (bytes): 198000000, relative offsets (bytes): 198000000, time_size (ms): 0.041888, time_decode (ms): 0.50864, achieved decode BW (GB/s): 778.547
MODE: normal,  RUNS_PER_THREAD: 1, DECODED_ITEMS_PER_THREAD: 3, THREADS_PER_BLOCK: 192, decoded size (bytes): 198000000, relative offsets (bytes): 0, time_size (ms): 0.031808, time_decode (ms): 0.336448, achieved decode BW (GB/s): 588.501
MODE: offsets, RUNS_PER_THREAD: 1, DECODED_ITEMS_PER_THREAD: 1, THREADS_PER_BLOCK: 128, decoded size (bytes): 198000000, relative offsets (bytes): 198000000, time_size (ms): 0.037472, time_decode (ms): 0.746208, achieved decode BW (GB/s): 530.683
MODE: normal,  RUNS_PER_THREAD: 1, DECODED_ITEMS_PER_THREAD: 1, THREADS_PER_BLOCK: 128, decoded size (bytes): 198000000, relative offsets (bytes): 0, time_size (ms): 0.04304, time_decode (ms): 0.611392, achieved decode BW (GB/s): 323.851
MODE: offsets, RUNS_PER_THREAD: 1, DECODED_ITEMS_PER_THREAD: 8, THREADS_PER_BLOCK: 128, decoded size (bytes): 198000000, relative offsets (bytes): 198000000, time_size (ms): 0.037792, time_decode (ms): 0.495168, achieved decode BW (GB/s): 799.729
MODE: normal,  RUNS_PER_THREAD: 1, DECODED_ITEMS_PER_THREAD: 8, THREADS_PER_BLOCK: 128, decoded size (bytes): 198000000, relative offsets (bytes): 0, time_size (ms): 0.039392, time_decode (ms): 0.2928, achieved decode BW (GB/s): 676.229
MODE: offsets, RUNS_PER_THREAD: 2, DECODED_ITEMS_PER_THREAD: 8, THREADS_PER_BLOCK: 128, decoded size (bytes): 198000000, relative offsets (bytes): 198000000, time_size (ms): 0.03232, time_decode (ms): 0.498016, achieved decode BW (GB/s): 795.155
MODE: normal,  RUNS_PER_THREAD: 2, DECODED_ITEMS_PER_THREAD: 8, THREADS_PER_BLOCK: 128, decoded size (bytes): 198000000, relative offsets (bytes): 0, time_size (ms): 0.039104, time_decode (ms): 0.298112, achieved decode BW (GB/s): 664.18
MODE: offsets, RUNS_PER_THREAD: 3, DECODED_ITEMS_PER_THREAD: 1, THREADS_PER_BLOCK: 256, decoded size (bytes): 198000000, relative offsets (bytes): 198000000, time_size (ms): 0.028, time_decode (ms): 1.18243, achieved decode BW (GB/s): 334.903
MODE: normal,  RUNS_PER_THREAD: 3, DECODED_ITEMS_PER_THREAD: 1, THREADS_PER_BLOCK: 256, decoded size (bytes): 198000000, relative offsets (bytes): 0, time_size (ms): 0.028768, time_decode (ms): 0.853088, achieved decode BW (GB/s): 232.098
MODE: offsets, RUNS_PER_THREAD: 1, DECODED_ITEMS_PER_THREAD: 8, THREADS_PER_BLOCK: 256, decoded size (bytes): 198000000, relative offsets (bytes): 198000000, time_size (ms): 0.046528, time_decode (ms): 0.49312, achieved decode BW (GB/s): 803.05
MODE: normal,  RUNS_PER_THREAD: 1, DECODED_ITEMS_PER_THREAD: 8, THREADS_PER_BLOCK: 256, decoded size (bytes): 198000000, relative offsets (bytes): 0, time_size (ms): 0.04064, time_decode (ms): 0.34272, achieved decode BW (GB/s): 577.731
MODE: offsets, RUNS_PER_THREAD: 8, DECODED_ITEMS_PER_THREAD: 1, THREADS_PER_BLOCK: 256, decoded size (bytes): 198000000, relative offsets (bytes): 198000000, time_size (ms): 0.030528, time_decode (ms): 1.84573, achieved decode BW (GB/s): 214.549
MODE: normal,  RUNS_PER_THREAD: 8, DECODED_ITEMS_PER_THREAD: 1, THREADS_PER_BLOCK: 256, decoded size (bytes): 198000000, relative offsets (bytes): 0, time_size (ms): 0.0288, time_decode (ms): 1.48307, achieved decode BW (GB/s): 133.507
MODE: offsets, RUNS_PER_THREAD: 1, DECODED_ITEMS_PER_THREAD: 1, THREADS_PER_BLOCK: 256, decoded size (bytes): 198000000, relative offsets (bytes): 198000000, time_size (ms): 0.0328, time_decode (ms): 0.967936, achieved decode BW (GB/s): 409.118
MODE: normal,  RUNS_PER_THREAD: 1, DECODED_ITEMS_PER_THREAD: 1, THREADS_PER_BLOCK: 256, decoded size (bytes): 198000000, relative offsets (bytes): 0, time_size (ms): 0.038304, time_decode (ms): 0.6552, achieved decode BW (GB/s): 302.198
MODE: offsets, RUNS_PER_THREAD: 2, DECODED_ITEMS_PER_THREAD: 2, THREADS_PER_BLOCK: 384, decoded size (bytes): 198000000, relative offsets (bytes): 198000000, time_size (ms): 0.02944, time_decode (ms): 0.870496, achieved decode BW (GB/s): 454.913
MODE: normal,  RUNS_PER_THREAD: 2, DECODED_ITEMS_PER_THREAD: 2, THREADS_PER_BLOCK: 384, decoded size (bytes): 198000000, relative offsets (bytes): 0, time_size (ms): 0.031328, time_decode (ms): 0.519552, achieved decode BW (GB/s): 381.098
```

## TODOs
- [ ] Describe how decoding into a window of the decode buffer works
- [ ] Performance optimisation: pack `uniques` and `offsets` into a single struct in shared memory if they both fit within a four-byte word.
- [ ] Performance optimisation: resurrect the `UnusedUnique`specialisation
- [ ] Split into three classes with common CRTP base class


allisonvacanti:
Thanks @elstehle! Just a heads up, I won't be able to look at this until I get back from vacation on 8/16, but @senior-zero will do an initial review in the meantime.

Do you need this merged for any particular release?

allisonvacanti:
I just noticed the TODOs -- should we wait until those are finished to start reviewing?

elstehle:
> Thanks @elstehle! Just a heads up, I won't be able to look at this until I get back from vacation on 8/16, but @senior-zero will do an initial review in the meantime.
> 
> Do you need this merged for any particular release?

Thanks, Allison. No rush. Enjoy your vacation. I'll align with @senior-zero in the meanwhile. 

> I just noticed the TODOs -- should we wait until those are finished to start reviewing?

I think, only the CRTP base class would be some more "intrusive" change (yet, not too intrusive). If we can get a decision on whether we want to pursue it, I'll push that. All other TODOs are either optional or really just minor additions.

cub/block/block_run_length_decode.cuh
```diff
@@ -0,0 +1,435 @@
...
+
+/// CUB namespace
+namespace cub
+{
+
+/**
+ * BlockRunLengthDecodeAlgorithm enumerates alternative algorithms for parallel
+ * reduction across a CUDA thread block.
+ */
+enum class BlockRunLengthDecodeAlgorithm
```
senior-zero, line 47: A few questions bother me:

- If both data and offset types are of the same size, this algorithm requires `2 * BLOCK_ITEMS` shared memory elements for unique items and run offsets and another `2 * BLOCK_ITEMS` for decode buffer. With only two items per thread, the occupancy is already limited by the shared memory size on P100. 

- To get the result, we have to iterate over block tiles. At each iteration, this algorithm has to perform a prefix scan and four thread-block synchronizations. Therefore, algorithm performance is strongly data-dependent. The slowdown should be a multiple of a decoded size divided by the block tile size. 

I've been thinking about an alternative solution. The main idea was that we only need two numbers per thread: value and an offset within the run. I've managed to write [an alternative algorithm ](https://gist.github.com/senior-zero/1d65beb94446cbec80c3c4f336a3f1b0) that requires only `2 * BLOCK_ITEMS` shared memory and doesn't have cooperative iterations after prefix scan computation. It's based on the merge path algorithm and merges prefix sum with the counting iterator. The binary search is performed in shared memory. The approach is similar to the `cub::DeviceSpmv`. The only difference is that instead of row prefix sum, I use runs prefix sums. And instead of a non-zero index, I use the index of an item in the decoded output. 

```cuda
// Prefix sum
int total_decoded_size {};

BlockScanT(temp_storage.scan)
  .InclusiveSum(lengths_reg, lengths_reg, total_decoded_size);
__syncthreads();

const int output_items_per_thread = (TILE_SIZE + total_decoded_size)
                                  / BLOCK_THREADS;

for (int i = 0; i < ITEMS; i++)
{
  int idx = static_cast<int>(threadIdx.x) * ITEMS + i;

  temp_storage.prefix[idx] = lengths_reg[i];
  shared_values[idx] = idx;
}
__syncthreads();

// Merge path
CoordinateT thread_coordinate;

{
  const int diagonal = output_items_per_thread
                     * static_cast<int>(threadIdx.x);

  cub::CountingInputIterator<int> nonzero_indices(0);

  MergePathSearch(diagonal,
                  temp_storage.prefix,
                  nonzero_indices,
                  TILE_SIZE,            // number of unique values (rows)
                  total_decoded_size,   // number of output elements (nnz)
                  thread_coordinate);
}

int value_id = thread_coordinate.x;
int offset = thread_coordinate.y;

int value = shared_values[value_id];
int value_end = temp_storage.prefix[value_id];

for (int i = 0; i < output_items_per_thread; i++)
{
  if (offset >= total_decoded_size)
  {
    break;
  }

  if (offset < value_end)
  {
    CONSUME(value, output); 

    offset++;
  }
  else
  {
    value_id++;
    value = shared_values[value_id];
    value_end = temp_storage.prefix[value_id];
  }
}
```

Preliminary benchmarks show that the speedup is about 3-4x. The speedup depends on the number of decoded items. 

The disadvantage of this approach is that the consecutive threads have items located further from each other. In your implementation, the stride is always equal to the items per thread (which is small). In the merge-path based approach, it's equal to the number of decoded items divided by thread block size, which might be any number.

I think we have two options. 
1) If there's no efficient way to reorder blocked data into the striped arrangement, we can expose this algorithm as a different API. Like `UNUSED`, `OFFSETS` and `NORMAL` implement different algorithms, we could add `MERGE_PATH` one.
2) If there's an efficient way to reorder blocked data into the striped arrangement - we could use only a merge path based approach. 

Both options require some work. I think we should proceed with this PR when the option is chosen.

cub/block/block_run_length_decode.cuh
```diff
@@ -0,0 +1,435 @@
...
+      , linear_tid(RowMajorTid(BLOCK_DIM_X, BLOCK_DIM_Y, BLOCK_DIM_Z))
+  {}
+
+  /**
+   * @brief Initializes the run-length decoding instances with the given sequence of runs and returns the total
+   * run-length decoded size. Subsequent calls to <b>RunLengthDecode</b> can be used to retrieve the run-length
+   * decoded data.
+   */
+  template <typename RunLengthT, typename OffsetT>
+  __device__ __forceinline__ void Init(UniqueItemT (&unique_items)[RUNS_PER_THREAD],
```
senior-zero, line 323: This method looks like a part of the constructor. We can't use other methods until this one is called.  I understand that this way, we can reuse the `BlockRunLengthDecode` object. But its non-Init constructor part is pretty lightweight, so there should be no performance issue in constructing a new one each time. Can we make this method private and call it from the constructor, or did I miss something?

cub/block/block_run_length_decode.cuh
```diff
@@ -0,0 +1,435 @@
...
+
+    // Ensure that the beginning of all relevant runs has been written
+    __syncthreads();
+
+    // Read in blocked-arrangement in preparation for the run-length decoding prefix scan
+    ZippedRunLengthDecode scan_items[DECODED_ITEMS_PER_THREAD];
+    LoadFromDecodeBuffer(scan_items, temp_storage.runs.decode_buffer);
+
+    // Ensure the decoded_buffer can be repurposed for the scan's temporary storage
+    __syncthreads();
```
senior-zero, line 421: For some reason, we use `CTA_SYNC()` macro instead of `__syncthraeds()` everywhere in CUB 😄

cub/block/block_run_length_decode.cuh
```diff
@@ -0,0 +1,435 @@
...
+   */
+  template <typename RunLengthT, typename OffsetT>
+  __device__ __forceinline__ void Init(UniqueItemT (&unique_items)[RUNS_PER_THREAD],
+                                       RunLengthT (&run_lengths)[RUNS_PER_THREAD],
+                                       OffsetT &total_decoded_size)
+  {
+    // Compute the offset for the beginning of each run
+    RelativeOffsetT run_offsets[RUNS_PER_THREAD];
+#pragma unroll
+    for (uint32_t i = 0; i < RUNS_PER_THREAD; i++)
```
senior-zero, line 330: `RUNS_PER_THREAD` is `int`, so we can't compare it with `uint32_t`. Can we change the type of `i` to `int` here and below?

cub/block/block_run_length_decode.cuh
```diff
@@ -0,0 +1,435 @@
...
+    // Ensure that the continuation-of-current-run items have been written to shared memory
+    __syncthreads();
+
+    // Scatter the beginning of each run into the run-length decode buffer
+    RunOffsetT run_offset = linear_tid;
+#pragma unroll
+    for (uint32_t i = 0; i < RUNS_PER_THREAD; i++)
+    {
+      // Fetch this thread's run and subsequently check whether it falls into this pass' run-length decode buffer
+      int32_t run_target_idx      = temp_storage.runs.run_offsets[run_offset];
```
senior-zero, line 388: The type of `temp_storage.runs.run_offsets` is `RelativeOffsetT`, can we change the type of `next_run_target_idx` to `RelativeOffsetT`?

cub/block/block_run_length_decode.cuh
```diff
@@ -0,0 +1,435 @@
...
+      // beginning of this buffer
+      bool is_a_trailing_run = (run_target_idx < from_decoded_offset) && (next_run_target_idx > from_decoded_offset);
+
+      // In particular for the trailing run, we want to scatter into index 0 (not a negative scatter index)
+      int32_t scatter_target_idx = max(0, run_target_idx - from_decoded_offset);
+      int32_t relative_offset    = is_a_trailing_run ? from_decoded_offset - run_target_idx : 0;
+
+      // Only write this beginning-of-run if the run length falls within the decoded buffer OR
+      // the run is a trailing run.
+      if (run_length > 0 && (run_target_idx >= from_decoded_offset || is_a_trailing_run) &&
```
senior-zero, line 402: We try to avoid using `&&` in a condition statement for flow control in CUB. Could you use nested if statements instead? The style guide is being developed 😄

senior-zero (changes requested):
It's a crucial algorithm to have in CUB! Thank you for developing it 😃 

I've proposed an alternative approach for RLE decode below. I think we should have a second iteration when it's considered.

cub/block/block_run_length_decode.cuh
elstehle: Thanks, I like the idea of using `MergePath` here! 👍  I think we still need to find a solution that will allow to retrieve a coalesced window of items and to iterate over those windows. But that should be doable with a `MergePath`-based approach.
I will also give a simple binary search a try, something like:
```
thread_out_offset = output_items_per_thread * threadIdx.x;
value_id = UpperBound(temp_storage.prefix, num_runs+1, thread_out_offset)-1];
```
That should reduce the binary search from `O(diagonal)` to `O(TILE_SIZE)`.

cub/block/block_run_length_decode.cuh
senior-zero: I think we can use something like [this](https://godbolt.org/z/r7TGErqG6). Or we can expose the PrefixProviderT as a template parameter and receive it in constructor :sweat_smile:

senior-zero: Anyway, I consider your idea of exposing an API that allows clients to provide us with an already computed prefix sum a must.

cub/block/block_run_length_decode.cuh
```diff
@@ -0,0 +1,435 @@
...
+#pragma once
+
+#include "../config.cuh"
+#include "../util_ptx.cuh"
+#include "../util_type.cuh"
+#include "block_scan.cuh"
+#include <limits>
+
+/// Optional outer namespace(s)
+CUB_NS_PREFIX
```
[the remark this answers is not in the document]
elstehle: Thanks! It's updated now 👍

cub/block/block_run_length_decode.cuh
elstehle: Thanks for the suggestion, @senior-zero! I switched to a binary search-based method. It simplified things quite a bit. No more need to have three different specialisations. We can get all the info we need from the binary search straight.
Would it still be worth to evaluate MergePath? Would we expect a benefit from MergePath over binary search?

senior-zero: I'm glad to hear that! 

> Would it still be worth to evaluate MergePath? Would we expect a benefit from MergePath over binary search?
I think that MergePath is a binary search in its essence. If your implementation of binary search doesn't duplicate it and you get similar performance improvements, there's no reason to force MergePath usage 😃

cub/block/block_run_length_decode.cuh
elstehle: Changed it to `CTA_SYNC()´ 👍

cub/block/block_run_length_decode.cuh
elstehle: Good catch. Fixed 👍

cub/block/block_run_length_decode.cuh
elstehle: Has become obsolete with the switch to binary search-based method.

cub/block/block_run_length_decode.cuh
elstehle: This has been removed with the switch to the binary search-based method. But I tried to adhere to it in the tests too 👍

cub/block/block_run_length_decode.cuh
elstehle: Have added the interface which allows the user to provide either (a) the array of the runs' lengths or (b) the array of the beginning-offsets of each run's length.

elstehle:
On top of addressing the review comments, I also switched to a compile-time-"unrolled" binary search that would always do log2(`NUMBER_OF_RUNS`) passes of binary search for all the threads, getting rid of some extra branching in the binary search. That had improved performance by another 5% on average.

`achieved BW STATIC` and `duration (ms) STATIC` represent the BW/time of the new binary search, that would have a statically determined number of binary search passes, whereas `achieved BW default` and `duration (ms) default` would denote the original binary search approach.

Experiments were run on a V100.

| RUNS_PER_THREAD | DECODED_ITEMS_PER_THREAD | THREADS_PER_BLOCK | time_decode STATIC | achieved BW STATIC | duration (ms) default | achieved BW default | speedup  |
|-----------------|--------------------------|-------------------|--------------------|--------------------|-----------------------|---------------------|----------|
|               1 |                        1 |                64 |            0.55936 |            722.254 |              0.601472 |             671.685 |  107.53% |
|               1 |                        1 |                64 |           0.487776 |            414.125 |              0.540544 |             373.698 |  110.82% |
|               1 |                        1 |                64 |           0.566144 |            713.599 |              0.604736 |              668.06 |  106.82% |
|               1 |                        1 |                64 |             0.4888 |            413.257 |              0.536896 |             376.237 |  109.84% |
|               1 |                        3 |               192 |           0.520352 |            776.398 |              0.518944 |             778.504 |   99.73% |
|               1 |                        3 |               192 |           0.301472 |            670.046 |              0.338432 |              596.87 |  112.26% |
|               1 |                        3 |               192 |           0.521152 |            775.206 |              0.528928 |             763.809 |  101.49% |
|               1 |                        3 |               192 |           0.289728 |            697.206 |               0.33712 |             599.193 |  116.36% |
|               1 |                        1 |               128 |           0.598688 |            674.809 |              0.667072 |             605.632 |  111.42% |
|               1 |                        1 |               128 |             0.5416 |            372.969 |              0.604192 |             334.331 |  111.56% |
|               1 |                        1 |               128 |           0.588256 |            686.776 |              0.650624 |             620.942 |  110.60% |
|               1 |                        1 |               128 |           0.537536 |            375.789 |              0.601728 |               335.7 |  111.94% |
|               1 |                        8 |               128 |            0.53088 |            761.001 |              0.523968 |             771.039 |   98.70% |
|               1 |                        8 |               128 |           0.282368 |            715.379 |              0.288256 |             700.766 |  102.09% |
|               1 |                        8 |               128 |           0.523008 |            772.455 |              0.513568 |             786.653 |   98.20% |
|               1 |                        8 |               128 |           0.284672 |            709.589 |              0.287648 |             702.247 |  101.05% |
|               2 |                        8 |               128 |           0.523648 |            771.511 |               0.52912 |             763.532 |  101.05% |
|               2 |                        8 |               128 |           0.288416 |            700.377 |                0.2904 |             695.592 |  100.69% |
|               2 |                        8 |               128 |           0.527136 |            766.406 |              0.527872 |             765.337 |  100.14% |
|               2 |                        8 |               128 |            0.28736 |            702.951 |              0.289472 |             697.822 |  100.74% |
|               3 |                        1 |               256 |            0.86096 |            469.244 |              0.845504 |             477.822 |   98.20% |
|               3 |                        1 |               256 |           0.739264 |            273.245 |              0.795808 |              253.83 |  107.65% |
|               3 |                        1 |               256 |           0.820992 |            492.088 |               0.87232 |             463.133 |  106.25% |
|               3 |                        1 |               256 |           0.746784 |            270.493 |              0.781408 |             258.508 |  104.64% |
|               1 |                        8 |               256 |           0.532384 |            758.851 |                0.5144 |             785.381 |   96.62% |
|               1 |                        8 |               256 |            0.28176 |            716.922 |              0.287712 |             702.091 |  102.11% |
|               1 |                        8 |               256 |           0.515712 |            783.383 |              0.517344 |             780.912 |  100.32% |
|               1 |                        8 |               256 |            0.30032 |            672.616 |              0.286784 |             704.363 |   95.49% |
|               8 |                        1 |               256 |            1.03683 |            389.648 |               1.09642 |             368.473 |  105.75% |
|               8 |                        1 |               256 |           0.941856 |             214.47 |               1.00131 |             201.735 |  106.31% |
|               8 |                        1 |               256 |             1.0384 |             389.06 |               1.10077 |             367.016 |  106.01% |
|               8 |                        1 |               256 |           0.945888 |            213.556 |              0.999488 |             202.103 |  105.67% |
|               1 |                        1 |               256 |           0.681152 |            593.113 |               0.73616 |             548.794 |  108.08% |
|               1 |                        1 |               256 |            0.59584 |            339.017 |              0.659936 |              306.09 |  110.76% |
|               1 |                        1 |               256 |           0.675616 |            597.973 |              0.741312 |              544.98 |  109.72% |
|               1 |                        1 |               256 |           0.588512 |            343.239 |              0.661696 |             305.276 |  112.44% |
|               2 |                        2 |               384 |            0.54528 |            740.904 |              0.554176 |              729.01 |  101.63% |
|               2 |                        2 |               384 |           0.426112 |            474.054 |              0.451648 |             447.251 |  105.99% |
|               2 |                        2 |               384 |           0.549056 |            735.808 |              0.557888 |              724.16 |  101.61% |
|               2 |                        2 |               384 |           0.430048 |            469.715 |              0.471104 |              428.78 |  109.55% |

PR #363, opened 2021-08-26
Merge sort docs and casts

gevtushenko:
This PR contains:
1) Documentation refactoring
2) Types clarifications
3) Direct usage of `CUB_NS_QUALIFIER` in the merge sort dispatch mechanism

### DIFF
--- a/cub/device/device_merge_sort.cuh
+++ b/cub/device/device_merge_sort.cuh
@@ -99,7 +99,7 @@ CUB_NAMESPACE_BEGIN
  *   CustomLess());
  * @endcode
  *
- * [LessThan Comparable] https://en.cppreference.com/w/cpp/named_req/LessThanComparable
+ * [LessThan Comparable]: https://en.cppreference.com/w/cpp/named_req/LessThanComparable
  */
 struct DeviceMergeSort
 {
@@ -131,8 +131,8 @@ struct DeviceMergeSort
    * CustomOpT custom_op;
    *
    * // Determine temporary device storage requirements
-   * void     *d_temp_storage = NULL;
-   * size_t   temp_storage_bytes = 0;
+   * void *d_temp_storage = nullptr;
+   * std::size_t temp_storage_bytes = 0;
    * cub::DeviceMergeSort::SortPairs(
    *   d_temp_storage, temp_storage_bytes,
    *   d_keys, d_values, num_items, custom_op);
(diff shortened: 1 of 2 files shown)


cub/device/device_merge_sort.cuh
```diff
@@ -126,45 +133,91 @@ struct DeviceMergeSort
...
+   *
+   * @tparam KeyIteratorT
+   *   is a model of [Random Access Iterator], `KeyIteratorT` is mutable, and
+   *   `KeyIteratorT`'s `value_type` is a model of [LessThan Comparable], and
+   *   the ordering relation on `KeyIteratorT`'s `value_type` is a
+   *   *strict weak ordering*, as defined in the [LessThan Comparable]
+   *   requirements
+   *
+   * @tparam ValueIteratorT
+   *   is a model of [Random Access Iterator], `ValueIteratorT` is mutable, and
```
zasdfgbnm, line 161: I think values don't need to be comparable?

cub/device/device_merge_sort.cuh
```diff
@@ -211,61 +267,117 @@ struct DeviceMergeSort
...
+   *   [LessThan Comparable], and the ordering relation on `KeyInputIteratorT`'s
+   *   `value_type` is a *strict weak ordering*, as defined in the
+   *   [LessThan Comparable] requirements
+   *
+   * @tparam ValueInputIteratorT
+   *   is a model of [Random Access Iterator], `ValueInputIteratorT` is mutable,
+   *   and `ValueInputIteratorT`'s `value_type` is a model of
+   *   [LessThan Comparable], and the ordering relation on
+   *   `ValueInputIteratorT`'s `value_type` is a *strict weak ordering*, as
+   *   defined in the [LessThan Comparable] requirements
```
zasdfgbnm, line 301: Same here, values don't need to be comparable?

cub/device/device_merge_sort.cuh
```diff
@@ -211,61 +267,117 @@ struct DeviceMergeSort
...
+   *   the ordering relation on `KeyIteratorT`'s `value_type` is a
+   *   *strict weak ordering*, as defined in the [LessThan Comparable]
+   *   requirements
+   *
+   * @tparam ValueIteratorT
+   *   is a model of [Random Access Iterator], `ValueIteratorT` is mutable, and
+   *   `ValueIteratorT`'s `value_type` is a model of [LessThan Comparable], and
+   *   the ordering relation on `ValueIteratorT`'s `value_type` is a
+   *   *strict weak ordering*, as defined in the [LessThan Comparable]
+   *   requirements
```
zasdfgbnm, line 315: Same here, values don't need to be comparable?

cub/device/device_merge_sort.cuh
```diff
@@ -474,46 +679,90 @@ struct DeviceMergeSort
...
+   *   and the ordering relation on `KeyIteratorT`'s `value_type` is a
+   *   *strict weak ordering*, as defined in the [LessThan Comparable]
+   *   requirements
+   *
+   * @tparam ValueIteratorT
+   *   is a model of [Random Access Iterator], `ValueIteratorT` is mutable, and
+   *   `ValueIteratorT`'s `value_type` is a model of [LessThan Comparable],
+   *   and the ordering relation on `ValueIteratorT`'s `value_type` is a
+   *   *strict weak ordering*, as defined in the [LessThan Comparable]
+   *   requirements
```
zasdfgbnm, line 710: Same here, values don't need to be comparable?

senior-zero: Thank you for this note! I've fixed everything from #372. The only difference is that I've left the requirement for `ValueIteratorT` being modifiable.

cub/device/dispatch/dispatch_merge_sort.cuh
```diff
@@ -555,21 +559,31 @@ struct DispatchMergeSort : SelectedPolicy
...
+
+      const auto temporary_keys_storage_size =
+        static_cast<std::size_t>(num_items * sizeof(KeyT));
+
+      const auto temporary_values_storage_size =
+        static_cast<std::size_t>(num_items * sizeof(ValueT)) * !KEYS_ONLY;
+
+      const auto virtual_shared_memory_size = vshmem_size(
+        static_cast<std::size_t>(max_shmem),
+        (CUB_NS_QUALIFIER::max)(block_sort_shmem_size, merge_shmem_size),
```
allisonvacanti, line 583: Just use `cub::` here -- the macro is only needed when used outside of the cub namespace. The other usages of `CUB_NS_MACRO` here can also just be `cub::`.

(The Thrust macro below is correct and should remain)

allisonvacanti (changes requested):

PR #374, opened 2021-09-13
[WIP] Allow cub::DeviceRadixSort and cub::DeviceSegmentedRadixSort to use iterator as input

zasdfgbnm:


cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1352,33 +1374,145 @@ struct DispatchRadixSort :
...
-                        DoubleBuffer<KeyT>(d_keys_tmp, d_keys_tmp2) :
-                        DoubleBuffer<KeyT>(d_keys_tmp2, d_keys_tmp);
-                    d_values = num_passes % 2 == 0 ?
-                        DoubleBuffer<ValueT>(d_values_tmp, d_values_tmp2) :
-                        DoubleBuffer<ValueT>(d_values_tmp2, d_values_tmp);
-                }
-                d_keys.selector ^= 1;
-                d_values.selector ^= 1;
+                input_mode = output_is_tmp ? TMP_STORAGE : OUTPUT;
+                output_is_tmp = !output_is_tmp;
```
zasdfgbnm, line 1515: @allisonvacanti Before I continue my work, I'd like to hear your feedback about this part. In order to support iterators, I have to add this verbose logic because the type of input iter, type of tmp storage, and type of output iter can be different.

zasdfgbnm: I am not sure if you are OK with this change, but this is very verbose, and I can not think of a better solution.

cub/device/device_radix_sort.cuh
```diff
@@ -179,46 +179,53 @@ struct DeviceRadixSort
...
-            d_keys,
-            d_values,
+            d_keys_in,
+            d_keys_out,
+            d_values_in,
+            d_values_out,
             num_items,
             begin_bit,
             end_bit,
-            is_overwrite_okay,
```
zasdfgbnm, line 221: `is_overwrite_okay` is removed, with iterator support, it never overwrites.

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1281,9 +1305,9 @@ struct DispatchRadixSort :
             // lookback
             max_num_blocks * RADIX_DIGITS * sizeof(AtomicOffsetT),
             // extra key buffer
-            is_overwrite_okay || num_passes <= 1 ? 0 : num_items * sizeof(KeyT),
+            num_passes <= 1 ? 0 : num_items * sizeof(KeyT),
```
allisonvacanti, line 1308: This will drastically increase the amount of temporary storage needed for some invocations of radix sort. We need to keep this optimization in place for the `DoubleBuffer` overloads in the `Device*RadixSort` APIs, since folks specifically use those to reduce the temporary storage allocations.

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1319,27 +1343,25 @@ struct DispatchRadixSort :
...
-                d_keys.d_buffers[1] = d_keys_tmp2;
-                d_values.d_buffers[1] = d_values_tmp2;
-            }
+
+            bool output_is_tmp = (num_passes % 2 == 0);
+            enum InputMode {
+                INPUT,
+                TMP_STORAGE,
+                OUTPUT
+            } input_mode = INPUT;
```
allisonvacanti, line 1364: Style: This should be split into separate declarations:

```
enum InputMode { ... };
InputMode input_mode = INPUT;
```

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1822,8 +1966,8 @@ struct DispatchSegmentedRadixSort :
             void* allocations[2] = {};
             size_t allocation_sizes[2] =
             {
-                (is_overwrite_okay) ? 0 : num_items * sizeof(KeyT),                      // bytes needed for 3rd keys buffer
-                (is_overwrite_okay || (KEYS_ONLY)) ? 0 : num_items * sizeof(ValueT),     // bytes needed for 3rd values buffer
+                num_items * sizeof(KeyT),                      // bytes needed for 3rd keys buffer
```
allisonvacanti, line 1969: Same as above, we need to keep this optimization in place.

allisonvacanti (changes requested):
Thanks for the patch! I'd like to get this functionality into CUB, but the proposed implementation will be too disruptive, since there is no way for a user to avoid the large temporary storage allocations needed to hold the intermediate keys/values. When users call the `DoubleBuffer` APIs, they are explicitly providing the scratch space needed to hold these intermediate results -- we need to use this memory instead of reallocating it.

I definitely want to merge this, but we'll need to preserve the `is_overwrite_okay` optimization and only allocate the extra scratch memory when it is absolutely necessary.

cub/device/dispatch/dispatch_radix_sort.cuh
allisonvacanti: We'll need to restore the `is_overwrite_okay` optimization first and then we can take a closer look at this part.

allisonvacanti:
Also -- @senior-zero is adding a new segmented sort implementation that uses iterator instead of pointers, see #357. It preserves the double buffer optimizations, so it may be a useful reference.

zasdfgbnm:
@allisonvacanti Looks like in #357, iterators are only supported for offsets, for input and output keys and values, they still need to be pointers?

zasdfgbnm:
1.16 is fine to me

cliffburdick:
Hi @zasdfgbnm , are there any updates on this?

PR #380, opened 2021-09-25
fix bugs for logical warp size is not a power of 2

YinLiu-91:
just as issue #179 said,when warp size is not a power of 2,it will give a incorrect anwer. It is because `laneId` computation is not correct when a logical warp is locatd in both physical warps when using shared memory reduction. For example, the logical warp size is 7,and there are 5 warps per block. In the fifth warp,it's thread id is 28 to 34,so it's lane id should be 0 to 6 accordingly , but former code will give a wrong laneId: 0,1,2,3,0,1,2,the correct is: 0,1,2,3,4,5,6.  Warp scan may also has the same problem

### DIFF
--- a/cub/warp/specializations/warp_reduce_smem.cuh
+++ b/cub/warp/specializations/warp_reduce_smem.cuh
@@ -113,7 +113,6 @@ struct WarpReduceSmem
 
         lane_id(IS_ARCH_WARP ?
             LaneId() :
-            threadIdx.x % LOGICAL_WARP_THREADS),
             (LaneId()+WARP_THREADS*WarpId()) % LOGICAL_WARP_THREADS),
             
 
--- a/cub/warp/specializations/warp_reduce_smem.cuh
+++ b/cub/warp/specializations/warp_reduce_smem.cuh
@@ -56,6 +56,7 @@ struct WarpReduceSmem
 
     enum
     {
+        WARP_THREADS = CUB_WARP_THREADS(PTX_ARCH),
         /// Whether the logical warp size and the PTX warp size coincide
         IS_ARCH_WARP = (LOGICAL_WARP_THREADS == CUB_WARP_THREADS(PTX_ARCH)),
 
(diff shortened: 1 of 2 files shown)


senior-zero:
Hello! Thank you for this work. Unfortunately, I don't think that it's a solution. First of all, if we partition a hardware warp into tiles of any non-power-of-two threads, we'll end up with the case below. 

image

It means that the warp synchronization used in the classes you've fixed won't actually synchronize threads of a virtual warp. Therefore after this fix, we have a data race. To avoid this issue, we'd have to use thread block synchronization, which indicates that this case is out of warp-scope layer. 

The second issue is the `WarpId` usage. As stated in the documentation:

> %warpid is volatile and returns the location of a thread at the moment when read, but its value may change during execution.

That means that we'd have to use thread id in the block. To compute it, we'd have to know thread block dimensions, which is another indication that this case is out of the warp-scope layer. 

Tests cover non-power-of-two modes for these classes.  The only difference is that in this mode, there's only one virtual warp instead of having multiple virtual warps per hardware warp. 

I'd say that the only place that's worth fixing is the documentation. It'd be better if we could explain this more clearly.

YinLiu-91:
Hello!
 Is non power of 2 case just work well  when `WARPS*LOGICAL_WARP_THREADS<LOGICAL_WARP_SIZES`? If so, it's just need to fix documents and add this restrictions in it. If it not be,  int the `test_warp_reduce.cu`, change codes like this:
```
/**
 * Run battery of tests for different number of active warps
 */
template <int LOGICAL_WARP_THREADS>
void Test()
{
     // Test<1, LOGICAL_WARP_THREADS>(); // changge 1 to 5
    Test<5, LOGICAL_WARP_THREADS>();

    // Only power-of-two subwarps can be tiled
    if ((LOGICAL_WARP_THREADS == 32) || PowerOfTwo<LOGICAL_WARP_THREADS>::VALUE)
        Test<2, LOGICAL_WARP_THREADS>();
}

```

It will give a `FAIL`  when `LOGICAL_WARP_THREADS=7` in 5 valid lanes case. This is the case when lane_id is wrong when logical warp crossing two hardware warps.

PR #386, opened 2021-10-21
Avoid include cycles with nvc++ -stdpar

dkolsen-pgi:
An `#include <algorithm>` in cub/device/dispatch/dispatch_scan_by_key.cuh
was causing compilation failures due to an include cycle when using
`nvc++ -stdpar`.  It doesn't look like `<algorithm>` is needed at all in
this file, so I am simply removing it.

### DIFF
--- a/cub/device/dispatch/dispatch_scan_by_key.cuh
+++ b/cub/device/dispatch/dispatch_scan_by_key.cuh
@@ -33,7 +33,6 @@
 #pragma once
 
 #include <iterator>
-#include <algorithm>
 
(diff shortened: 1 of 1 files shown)


dkolsen-pgi:
If it is discovered during testing that `#include <algorithm>` is really necessary, then it needs to be changed to `#include <thrust/detail/algorithm_wrapper.h>` which avoids the include_cycle problem.

allisonvacanti:
DVS CL: 30563500
gpuCI: NVIDIA/thrust#1542

PR #400, opened 2021-10-29
Optimize compilation time for the common case

senior-zero:
This PR contains compilation time improvement for the common case when agents fit into default shared memory size (48 KB).

### DIFF
--- a/cub/device/dispatch/dispatch_merge_sort.cuh
+++ b/cub/device/dispatch/dispatch_merge_sort.cuh
@@ -27,10 +27,10 @@
 
 #pragma once
 
-#include "../../util_math.cuh"
-#include "../../util_device.cuh"
-#include "../../util_namespace.cuh"
-#include "../../agent/agent_merge_sort.cuh"
+#include <cub/util_math.cuh>
+#include <cub/util_device.cuh>
+#include <cub/util_namespace.cuh>
+#include <cub/agent/agent_merge_sort.cuh>
 
 #include <thrust/system/cuda/detail/core/triple_chevron_launch.h>
 #include <thrust/detail/integer_math.h>
@@ -245,7 +245,8 @@ template <typename KeyInputIteratorT,
           typename ActivePolicyT,
           typename CompareOpT,
           typename KeyT,
-          typename ValueT>
+          typename ValueT,
+          bool AgentFitsIntoDefaultShmemSize>
 struct BlockSortLauncher
 {
   int num_tiles;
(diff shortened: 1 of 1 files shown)


allisonvacanti (approved):
Thanks @senior-zero, this is a huge improvement -- I'm seeing compilation time improve on both nvcc and nvc++. It isn't quite back to before, but it's a significant reduction!

Compile times for the `thrust::sort` test program:

| Compiler | Old merge sort | Current merge sort | This PR |
|-|-|-|-|
| nvcc | 18.79s | 29.10s (+55%) | 22.44s (+19%) |
| nvc++ | 61.81s | 75.61s (+22%) | 65.80s (+6%) |

LGTM -- Let's get this merged once the tests are passing and see how this impacts the total build time.

allisonvacanti:
Related to NVBugs 3418930 and 3419768.

dongxiao92:
Could you help to explain why these changes can reduce compilation time?
From the code changes, I understand that 1) instantiation for DeviceMergeSortBlockSortKernel with use_vshmem=true may not be needed and 2) compilation for code in [this](https://github.com/NVIDIA/cub/pull/400/files#diff-94c2c8edc681c63eddebe5e8fa3bb312ba83aabe9bc3dc6bf0e2808bdef9dbbeR584) if branch may not be needed.
Does the compilation time reduction come from these two changes?

senior-zero:
> Could you help to explain why these changes can reduce compilation time? From the code changes, I understand that 1) instantiation for DeviceMergeSortBlockSortKernel with use_vshmem=true may not be needed and 2) compilation for code in [this](https://github.com/NVIDIA/cub/pull/400/files#diff-94c2c8edc681c63eddebe5e8fa3bb312ba83aabe9bc3dc6bf0e2808bdef9dbbeR584) if branch may not be needed. Does the compilation time reduction come from these two changes?

Hello, @dongxiao92! 

Two specializations of merge sort kernels exist:
| kernel                         | use vshmem=false | use vshmem=true |
|--------------------------------|------------------|-----------------|
| DeviceMergeSortBlockSortKernel | +                | +               |
| DeviceMergeSortMergeKernel     | +                | +               |

Since the check for available shared memory is performed at runtime, we had to compile for both cases (in generic case). This patch relies on the thrust approach which consists of comparing kernel shared memory size requirements with the default available shared memory size (48KB). This check can be done at compile time. If we know that virtual shared memory is not required at compile-time, there's no need to compile merge sort kernels twice.

Issue #295, opened 2021-05-05
Failure when running a float to float reduction


zkhatami:
Two tests fail when running a float to float reduction:
1) cub.test.device_scan.minimal
2) cub.test.device_reduce.minimal

The failure occurs when running a float to float scan/reduction:

1) For cub.test.device_scan.minimal:
The failure is not related to NVC++ nor CUB. It's actually in the test itself, where the value of h_reference[i] gets calculated within a "Solve" call. Test fails when comparing the result correctness to 9 significant figures and it fails with:

INCORRECT: [372510]: (computed) 6.93817e+10 != 6.93747e+10 (difference:6.98778e+06, fraction: 0.000100725)

The naïve approach of summing up the floating numbers, the one that "Solve" call is currently doing, results in having the accuracy of output to less significant figures:

for (int i = 1; i < num_items; ++i)
{
val = static_cast<AccumT>(h_in[i]);
h_reference[i] = static_cast<OutputT>(inclusive);
inclusive = scan_op(inclusive, val);
}

I used Kahan’s method instead which is as accurate as possible using 32-bit numbers:

OutputT s = h_in[0];
OutputT c = 0.0;
OutputT y = 0.0;
OutputT t = 0.0;

for (int i = 1; i < num_items; ++i) {
h_reference[i] = s;
y = h_in[i] - c;
t = s + y;
c = (t - s) - y;
s = t;
}

and test gets passed.

2) For cub.test.device_reduce.minimal, the reduce function is sum() and if we use Kahan's algorithm, we can get more accurate result to 9 significant figures when using float type input that makes this test to be passed.

I have attached simple example that shows the accuracy difference between using naïve sum and Kahan's sum and it shows that Kahan results with the same accuracy as double inputs:

naive sum result on float = 14.3573579788208
Kahan's sum result on float = 14.3927268981934
naive sum result on double = 14.3927268981934

[float_sum.txt](https://github.com/NVIDIA/cub/files/6429704/float_sum.txt)

allisonvacanti:
PR #402 should address this. I couldn't reproduce this on nvc++ 21.7 -- if you get a chance to check that this fixes the failing nvc++ tests, let me know!

> I used Kahan’s method instead

Unfortunately we can't really use this technique, since `scan_op` may be a different operation than `cub::Sum`. Since the error you're seeing (0.0100725%) is only slightly more than the current threshold (0.001%), I've just increased the threshold to (0.0015%).

PR #402, opened 2021-11-08
Increase floating point tolerance when validating results.

allisonvacanti:
#295 reported that the `device_reduce` and `device_scan` tests were
failing due to the accumulated fp error exceeding the current threshold
of 0.01%. They observed a relative error of slightly more, 0.0100725%.

This appears to be happening on nvc++, though I've not been able to repro
the failure on NV HPC 21.7.

They suggest an alternate method of computing the summation that provides
more accuracy, but unfortunately this approach is not suitable for testing
cases where `scan_op` is not `cub::Sum`.

Instead, let's just bump up the error threshold to 0.015%.

### DIFF
--- a/test/test_util.h
+++ b/test/test_util.h
@@ -1428,7 +1428,7 @@ int CompareResults(float* computed, float* reference, OffsetT len, bool verbose 
             float difference = std::abs(computed[i]-reference[i]);
             float fraction = difference / std::abs(reference[i]);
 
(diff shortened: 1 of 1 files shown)


allisonvacanti:
gpuCI: NVIDIA/thrust#1562

allisonvacanti:
run tests

senior-zero (approved):

PR #404, opened 2021-11-10
Fix bad cast in util_device.

### LINKED ISSUE #286: Illegal function pointer to void* cast.
https://github.com/NVIDIA/cub/blob/main/cub/util_device.cuh#L369

cudaFuncGetAttribute's 2nd parameter is const void*. Implicit casts from pointers to functions to void* are disallowed. You need an implicit cast.

```cpp
void func(const void*) { }

int main() {
  typedef void (*EmptyKernelPtr)();
  EmptyKernelPtr empty_kernel;

  func(empty_kernel);
}
```
```
call.cxx:7:3: error: no matching function for call to 'func'
  func(empty_kernel);
  ^~~~
call.cxx:1:6: note: candidate function not viable: no known conversion from 'EmptyKernelPtr' (aka 'void (*)()') to 'const void *' for 1st argument; take the address of the argument with &
void func(const void*) { }
     ^
```

I don't know why this compiles currently.

allisonvacanti:
Fixes #286.

### DIFF
--- a/cub/util_device.cuh
+++ b/cub/util_device.cuh
@@ -371,11 +371,9 @@ CUB_RUNTIME_FUNCTION inline cudaError_t PtxVersionUncached(int& ptx_version)
        #if CUB_INCLUDE_HOST_CODE
             cudaFuncAttributes empty_kernel_attrs;
 
-            do {
-                if (CubDebug(result = cudaFuncGetAttributes(&empty_kernel_attrs, empty_kernel)))
-                    break;
-            }
-            while(0);
+            result = cudaFuncGetAttributes(&empty_kernel_attrs,
+                                           reinterpret_cast<void*>(empty_kernel));
+            CubDebug(result);
 
             ptx_version = empty_kernel_attrs.ptxVersion * 10;
         #endif


allisonvacanti:
DVS CL: 30637405
gpuCI: NVIDIA/thrust#1567

cub/util_device.cuh
```diff
@@ -364,7 +364,7 @@ CUB_RUNTIME_FUNCTION inline cudaError_t PtxVersionUncached(int& ptx_version)
 
     // This is necessary for unused variable warnings in host compilers. The
     // usual syntax of (void)empty_kernel; was not sufficient on MSVC2015.
-    (void)reinterpret_cast<void*>(empty_kernel);
+    (void)reinterpret_cast<void*>(&empty_kernel);
```
senior-zero, line 367: As I understand the original issue is related to:

```cuda
cudaFuncGetAttributes(&empty_kernel_attrs, empty_kernel))
// -->
cudaFuncGetAttributes(&empty_kernel_attrs, reinterpret_cast<void*>(empty_kernel)))
```

Here's [the reproducer](https://cuda.godbolt.org/z/zshYzfYv4). Somehow `cudaFuncGetAttributes` doesn't emit the warning. 

Has `(void)reinterpret_cast<void*>(&empty_kernel);` fixed any warnings in your case? If not, I suggest we leave this line as it is and cast the pointer explicitly in `cudaFuncGetAttributes`.

senior-zero (changes requested):

cub/util_device.cuh
allisonvacanti: Ah, the line number have changed since that bug was reported, and the link is currently pointing to the `reinterpret_cast`. Thanks for spotting this!

I'll update this.

allisonvacanti:
DVS CL: 30643838
gpuCI: NVIDIA/thrust#1567

senior-zero (approved):

PR #410, opened 2021-12-03
Fix segmented sort device-side launch

### LINKED ISSUE #409: DeviceSegmentedSort synchronizes default stream and produces wrong results when launched from a kernel
1. DeviceSegmentedSort performs cudaMemcpy on the default stream instead of the supplied stream.
2. In device code, memcpy may execute before group_sizes and num_selected_groups are calculated. This leads to wrong results.

```
    if (CUB_IS_HOST_CODE)
    {
      #if CUB_INCLUDE_HOST_CODE
      if (CubDebug(error = cudaMemcpy(h_group_sizes,
                                      group_sizes.get(),
                                      num_selected_groups * sizeof(unsigned int),
                                      cudaMemcpyDeviceToHost)))
      {
        return error;
      }
      #endif
    }
    else
    {
      #if CUB_INCLUDE_DEVICE_CODE
      memcpy(h_group_sizes,
             group_sizes.get(),
             num_selected_groups * sizeof(unsigned int));
      #endif
    }
```

senior-zero:
This PR fixes #409: 
1. Host-side calls don't synchronize with the default stream.
2. Device-side calls synchronize stream before `group_sizes` usage.

I've also added a simple device-side launch test. In order to get it compiling, run CMake with
```
cmake -DCUB_ENABLE_TESTS_WITH_RDC=ON ...
```

and add `#define CUB_CDP` in the test_device_segmented_sort.cu.

### DIFF
--- a/cub/device/dispatch/dispatch_segmented_sort.cuh
+++ b/cub/device/dispatch/dispatch_segmented_sort.cuh
@@ -1386,7 +1386,7 @@ private:
         return error;
       }
 
-      if (CubDebug(error = cudaStreamSynchronize(stream)))
+      if (CubDebug(error = SyncStream(stream)))
       {
         return error;
       }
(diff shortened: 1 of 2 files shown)


allisonvacanti (approved):

senior-zero:
gpuCI: NVIDIA/thrust#1576

PR #411, opened 2021-12-03
Only use nv_exec_check_disable on nvcc.

allisonvacanti:
This patch adds the same guards around the pragma that we use in Thrust.

Bug 3453249

### DIFF
--- a/cub/detail/exec_check_disable.cuh
+++ b/cub/detail/exec_check_disable.cuh
@@ -1,0 +1,41 @@
+/*
+*  Copyright 2021 NVIDIA Corporation
+*
+*  Licensed under the Apache License, Version 2.0 (the "License");
+*  you may not use this file except in compliance with the License.
+*  You may obtain a copy of the License at
+*
+*      http://www.apache.org/licenses/LICENSE-2.0
+*
+*  Unless required by applicable law or agreed to in writing, software
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*/
+
+#pragma once
+
+#include <cub/util_compiler.cuh>
+
+/**
+ * @def CUB_EXEC_CHECK_DISABLE
+ * Wrapper around `#pragma nv_exec_check_disable`.
+ */
+
+// #pragma nv_exec_check_disable is only recognized by NVCC.
+#if defined(__CUDACC__) && \
+    !defined(_NVHPC_CUDA) && \
+    !(defined(__CUDA__) && defined(__clang__))
+
+#if CUB_HOST_COMPILER == CUB_HOST_COMPILER_MSVC
+#define CUB_EXEC_CHECK_DISABLE __pragma("nv_exec_check_disable")
+#else // // !MSVC
+#define CUB_EXEC_CHECK_DISABLE _Pragma("nv_exec_check_disable")
+#endif // MSVC
+
+#else // !NVCC
+
+#define CUB_EXEC_CHECK_DISABLE
+
+#endif // NVCC
--- a/cub/detail/device_synchronize.cuh
+++ b/cub/detail/device_synchronize.cuh
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <cub/detail/exec_check_disable.cuh>
 #include <cub/util_arch.cuh>
 #include <cub/util_namespace.cuh>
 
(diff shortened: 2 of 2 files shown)


allisonvacanti:
gpuCI: NVIDIA/thrust#1574
DVS CL: 30734665

cub/detail/exec_check_disable.cuh
```diff
@@ -0,0 +1,41 @@
...
+#include <cub/util_compiler.cuh>
+
+/**
+ * \def CUB_EXEC_CHECK_DISABLE
+ * Wrapper around `#pragma nv_exec_check_disable`.
+ */
+
+// #pragma nv_exec_check_disable is only recognized by NVCC.
+#if defined(__CUDACC__) && \
+    !defined(_NVHPC_CUDA) && \
```
senior-zero, line 28: As I understand `_NVHPC_CUDA` is a replacement for `__NVCOMPILER_CUDA__`. Since CUB uses both, it's a bit confusing. Are you going to replace `__NVCOMPILER_CUDA__` in the following PRs?

cub/detail/exec_check_disable.cuh
```diff
@@ -0,0 +1,41 @@
...
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*/
+
+#pragma once
+
+#include <cub/util_compiler.cuh>
+
+/**
+ * \def CUB_EXEC_CHECK_DISABLE
```
senior-zero, line 22: Use `@def` for consistency.

cub/detail/exec_check_disable.cuh
```diff
@@ -0,0 +1,41 @@
...
+ * Wrapper around `#pragma nv_exec_check_disable`.
+ */
+
+// #pragma nv_exec_check_disable is only recognized by NVCC.
+#if defined(__CUDACC__) && \
```
senior-zero, line 27: Nitpick: I'd like to see something like:

```cuda
#if CUB_DEVICE_COMPILER == CUB_DEVICE_COMPILER_NVCC
/// ... 
#endif 
```

But the `CUB_DEVICE_COMPILER` macro seems to have a different meaning. It uses NVCC as a device compiler even in the case of clang usage. I haven't found any usages of `CUB_DEVICE_COMPILER`. If you have some time to repurpose it to actually define which compiler is used for device code, it would be possible to rewrite this macro as shown above. 

If this is a hotfix, feel free to ignore this suggestion and we'll clean this up later.

cub/detail/exec_check_disable.cuh
```diff
@@ -0,0 +1,41 @@
...
+    !defined(_NVHPC_CUDA) && \
+    !(defined(__CUDA__) && defined(__clang__))
+
+#if CUB_HOST_COMPILER == CUB_HOST_COMPILER_MSVC
+#define CUB_EXEC_CHECK_DISABLE __pragma("nv_exec_check_disable")
+#else // MSVC
```
senior-zero, line 33: Nitpick: as I understand, it's common to keep else macro undocumented or write opposite condition in the comment:

```cuda
#if A 
#else // !A
#endif 
```

Without reading the condition, it seems like the else case is for the MSVC case, but it isn't. The same comment can be made for the code below:

```cuda
#else // NVCC
// Non-NVCC case
#endif // NVCC
```

cub/detail/exec_check_disable.cuh
```diff
@@ -0,0 +1,41 @@
...
+ * Wrapper around `#pragma nv_exec_check_disable`.
+ */
+
+// #pragma nv_exec_check_disable is only recognized by NVCC.
+#if defined(__CUDACC__) && \
+    !defined(_NVHPC_CUDA) && \
+    !(defined(__CUDA__) && defined(__clang__))
+
+#if CUB_HOST_COMPILER == CUB_HOST_COMPILER_MSVC
+#define CUB_EXEC_CHECK_DISABLE __pragma("nv_exec_check_disable")
```
senior-zero, line 32: There's a bug 3423950 related to `__pragma` usage in nvcc 11.5 with `nv_diagnostic`. It leads to an internal compilation error on windows. I'm not sure if it's `nv_diagnostic` specific or not. Can you check this, please? If this issue is related to `__pragma` in general, we'll have to add another condition in the compiler check.

senior-zero (changes requested):

cub/detail/exec_check_disable.cuh
allisonvacanti: `_NVHPC_CUDA` is preferred in the future -- the other one will not be around much longer AIUI

cub/detail/exec_check_disable.cuh
allisonvacanti: Following up on our conversation, we'll clean this up in a future PR.

cub/detail/exec_check_disable.cuh
allisonvacanti: We've tested `__pragma("nv_exec_check_disable")` in Thrust for the last several releases (including 11.5), it shouldn't have the same issue as 3423950.

PR #421, opened 2022-01-17
Fix spurious uninitialized variable warning on gcc11.

allisonvacanti:
Replace the type-punning with a UB-safe bit conversion. The
compiler will optimize the memcpy out and just convert the bits.


allisonvacanti:
gpuCI: NVIDIA/thrust#1590
DVS CL: 30883617

PR #424, opened 2022-01-19
Fix segfault in device radix sort tests

senior-zero:
There was a case where we were testing `num_items = 0 && num_segments = 1`. The radix sort tests don't support this, which leads to out of bounds accesses in multiple places. Somehow this issue doesn't show itself on Linux. All tests of device radix sort were failing on Windows in a racy way.

### DIFF
--- a/test/test_device_radix_sort.cu
+++ b/test/test_device_radix_sort.cu
@@ -1242,9 +1242,19 @@ void TestSegments(
     }
 
     // Test single segment
-    if (num_items < 128 * 1000 || pre_sorted) {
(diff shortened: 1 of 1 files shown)


allisonvacanti (approved):

allisonvacanti:
You can skip DVS testing for this one, since it doesn't run the CUB tests. Just gpuCI will be fine.

allisonvacanti:
Rebased to pull in CI fixes.

allisonvacanti:
gpuCI: NVIDIA/thrust#1595
DVS: Not needed (CUB unit test changes only)

PR #426, opened 2022-01-21
Fix version checks in CMake packages.

allisonvacanti:


allisonvacanti:
gpuCI: NVIDIA/thrust#1599

robertmaynard (approved):

Issue #406, opened 2021-11-15
cub::DeviceSelect::Unique doesn't work with thrust::discard_iterator


zasdfgbnm:
Sample code:
```c++
#include <cub/cub.cuh>
#include <thrust/iterator/discard_iterator.h>

int main() {
    int *p = nullptr;
    thrust::discard_iterator<int> d;
    size_t s;

    cub::DeviceSelect::Unique(nullptr, s, p, p, p, 0);    // works
    cub::DeviceSelect::Unique(nullptr, s, p, p, d, 0);    // works
    // cub::DeviceSelect::Unique(nullptr, s, p, d, p, 0);    // Error
    // cub::DeviceSelect::Unique(nullptr, s, p, d, d, 0);    // Error
}
```

https://github.com/NVIDIA/cub/pull/405 is also affected. I am not fixing this issue in https://github.com/NVIDIA/cub/pull/405, so the fix of this should fix `cub::DeviceSelect::UniqueByKey` as well.

allisonvacanti:
More info: https://www.godbolt.org/z/8q9qTcxrK

```
/opt/compiler-explorer/libs/thrustcub/1.13.0/cub/block/specializations/../../block/../thread/../thread/thread_operators.cuh(64): error: no operator "==" matches these operands
            operand types are: const thrust::detail::any_assign == const thrust::detail::any_assign
          detected during:
            instantiation of "__nv_bool cub::Equality::operator()(const T &, const T &) const [with T=thrust::detail::any_assign]" 
(100): here
            instantiation of "__nv_bool cub::InequalityWrapper<EqualityOp>::operator()(const T &, const T &) [with EqualityOp=cub::Equality, T=thrust::detail::any_assign]" 
/opt/compiler-explorer/libs/thrustcub/1.13.0/cub/block/specializations/../../block/block_discontinuity.cuh(162): here
            instantiation of "__nv_bool cub::BlockDiscontinuity<T, BLOCK_DIM_X, BLOCK_DIM_Y, BLOCK_DIM_Z, PTX_ARCH>::ApplyOp<FlagOp, false>::FlagT(FlagOp, const T &, const T &, int) [with T=thrust::detail::any_assign, BLOCK_DIM_X=128, BLOCK_DIM_Y=1, BLOCK_DIM_Z=1, PTX_ARCH=520, FlagOp=cub::InequalityWrapper<cub::Equality>]" 
[...]

PR #444, opened 2022-03-16
Fix select if for mixed types

senior-zero:
During three-way partitioning development I've noticed that there's a bug in the select if implementation. It used output type to load data from input iterator. Recently, @elstehle brought the very same issue. Here's a code sample that fails to compile. This fact limits the applicability of the partitioning/selection facilities:

```cpp
#include <iostream>

#include <thrust/count.h>
#include <thrust/iterator/zip_iterator.h>
#include <thrust/iterator/transform_output_iterator.h>
#include <thrust/iterator/discard_iterator.h>
#include <thrust/device_vector.h>

#include <cub/device/device_select.cuh>

struct pair_to_col {
    __host__ __device__ int operator()(const thrust::tuple<int, double> &in) {
        return thrust::get<0>(in);
    }
};

struct select {
    __host__ __device__ bool operator()(const thrust::tuple<int, double> &in) {
        return static_cast<double>(thrust::get<0>(in)) > thrust::get<1>(in);
    }
};

int main() {
    const std::size_t n = 1024;
    thrust::device_vector<int> col_a(n, 42);
    thrust::device_vector<double> col_b(n, 4.2);

    thrust::device_vector<int> result(n, 42);

    auto in = thrust::make_zip_iterator(col_a.begin(), col_b.begin());
    auto out = thrust::make_transform_output_iterator(result.begin(), pair_to_col{});

    std::size_t tmp_storage_size{};
    cub::DeviceSelect::If(
            nullptr, tmp_storage_size, in, out, thrust::make_discard_iterator(), n, select{});

    thrust::device_vector<char> tmp_storage(tmp_storage_size);
    cub::DeviceSelect::If(
            thrust::raw_pointer_cast(tmp_storage.data()), tmp_storage_size, in, out, thrust::make_discard_iterator(), n, select{});

    if (thrust::count(result.begin(), result.end(), 42) == n) {
        std::cout << "OK" << std::endl;
    } else {
        std::cout << "FAIL" << std::endl;
    }

    return 0;
}
```  

This PR addresses this issue, allowing different value types for input and output iterators. It's worth mentioning that the implementation used to have correct types up to the following [commit](47659788537a2727fd0f3452243696a845015fa8) which doesn't contain any mentions of the value types in its description. So my guess whould be that it was a typo rather than an intent. 

I've also noticed that the radix sort test fails to compile on windows. I've fixed a few conversion issues here.

### DIFF
--- a/cub/device/dispatch/dispatch_radix_sort.cuh
+++ b/cub/device/dispatch/dispatch_radix_sort.cuh
@@ -1278,9 +1278,11 @@ struct DispatchRadixSort :
         const PortionOffsetT PORTION_SIZE = ((1 << 28) - 1) / ONESWEEP_TILE_ITEMS * ONESWEEP_TILE_ITEMS;
         int num_passes = cub::DivideAndRoundUp(end_bit - begin_bit, RADIX_BITS);
         OffsetT num_portions = static_cast<OffsetT>(cub::DivideAndRoundUp(num_items, PORTION_SIZE));
-        PortionOffsetT max_num_blocks = cub::DivideAndRoundUp(CUB_MIN(num_items, PORTION_SIZE),
-                                                           ONESWEEP_TILE_ITEMS);
-        
+        PortionOffsetT max_num_blocks = cub::DivideAndRoundUp(
+          static_cast<int>(
+            CUB_MIN(num_items, static_cast<OffsetT>(PORTION_SIZE))),
+          ONESWEEP_TILE_ITEMS);
+
         size_t value_size = KEYS_ONLY ? 0 : sizeof(ValueT);
         size_t allocation_sizes[] =
         {
@@ -1355,7 +1357,10 @@ struct DispatchRadixSort :
                 int num_bits = CUB_MIN(end_bit - current_bit, RADIX_BITS);
                 for (OffsetT portion = 0; portion < num_portions; ++portion)
                 {
-                    PortionOffsetT portion_num_items = CUB_MIN(num_items - portion * PORTION_SIZE, PORTION_SIZE);
+                    PortionOffsetT portion_num_items =
+                      static_cast<PortionOffsetT>(
+                        CUB_MIN(num_items - portion * PORTION_SIZE,
+                                static_cast<OffsetT>(PORTION_SIZE)));
                     PortionOffsetT num_blocks =
                         cub::DivideAndRoundUp(portion_num_items, ONESWEEP_TILE_ITEMS);
                     if (CubDebug(error = cudaMemsetAsync(
--- a/test/test_device_radix_sort.cu
+++ b/test/test_device_radix_sort.cu
@@ -35,6 +35,7 @@
 
 #include <algorithm>
 #include <cstdio>
+#include <limits>
 #include <memory>
 #include <random>
 #include <type_traits>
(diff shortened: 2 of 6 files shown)


allisonvacanti:
@senior-zero Sounds like this may also fix #406, can you confirm?

elstehle (approved):
Thanks for jumping onto this so quickly, @senior-zero! LGTM. This will definitely help my use case. Given that the change also follows the behaviour we have in the prefix scan, I think it makes sense too. I'm curious if, originally, making `InputIt::value_type` the `internal_type` was a conscious decision and if there's rationale behind.

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1355,7 +1356,8 @@ struct DispatchRadixSort :
                 int num_bits = CUB_MIN(end_bit - current_bit, RADIX_BITS);
                 for (OffsetT portion = 0; portion < num_portions; ++portion)
                 {
-                    PortionOffsetT portion_num_items = CUB_MIN(num_items - portion * PORTION_SIZE, PORTION_SIZE);
+                    PortionOffsetT portion_num_items =
+                            CUB_MIN(static_cast<PortionOffsetT>(num_items - portion * PORTION_SIZE), PORTION_SIZE);
```
allisonvacanti, line 1360: Same as above.

test/test_device_radix_sort.cu
```diff
@@ -285,7 +285,7 @@ cudaError_t Dispatch(
 {
     return DeviceSegmentedRadixSort::SortPairs(
         d_temp_storage, temp_storage_bytes,
-        d_keys, d_values, num_items,
+        d_keys, d_values, static_cast<int>(num_items),
```
allisonvacanti, line 288: Add a runtime check that fails the test if `num_items` doesn't fit into an `int`, since this test file also covers the `DeviceRadixSort` 64-bit offsets and we may end up passing larger `num_items` here.

test/test_device_radix_sort.cu
```diff
@@ -323,7 +323,7 @@ cudaError_t Dispatch(
     cudaError_t retval = DeviceSegmentedRadixSort::SortPairs(
         d_temp_storage, temp_storage_bytes,
         const_keys_itr, d_keys.Alternate(), const_values_itr, d_values.Alternate(),
-        num_items, num_segments, d_segment_begin_offsets, d_segment_end_offsets,
+        static_cast<int>(num_items), num_segments, d_segment_begin_offsets, d_segment_end_offsets,
```
allisonvacanti, line 326: Same as above.

allisonvacanti (changes requested):

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1278,8 +1278,9 @@ struct DispatchRadixSort :
         const PortionOffsetT PORTION_SIZE = ((1 << 28) - 1) / ONESWEEP_TILE_ITEMS * ONESWEEP_TILE_ITEMS;
         int num_passes = cub::DivideAndRoundUp(end_bit - begin_bit, RADIX_BITS);
         OffsetT num_portions = static_cast<OffsetT>(cub::DivideAndRoundUp(num_items, PORTION_SIZE));
-        PortionOffsetT max_num_blocks = cub::DivideAndRoundUp(CUB_MIN(num_items, PORTION_SIZE),
-                                                           ONESWEEP_TILE_ITEMS);
+        PortionOffsetT max_num_blocks = cub::DivideAndRoundUp(
+                CUB_MIN(static_cast<PortionOffsetT>(num_items), PORTION_SIZE),
```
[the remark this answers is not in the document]
senior-zero: Good point, I've missed this part. Thanks!

test/test_device_select_if.cu
```diff
@@ -652,6 +658,74 @@ void Test(
     }
 }
 
+template<class FT, class ST>
```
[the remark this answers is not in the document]
senior-zero: I'm sorry for the confusion 😄 `T0` and `T1` do look less ambiguous, thanks!

allisonvacanti:
gpuCI: NVIDIA/thrust#1633
DVS CL: 31093334

PR #465, opened 2022-04-27
Remove unused/obsolete files.

allisonvacanti:
These files are not used or distributed, and often use API that's no longer present in the current repo.


allisonvacanti:
gpuCI: NVIDIA/thrust#1673

senior-zero (approved):

PR #466, opened 2022-04-28
Alternative approach to 64-bit indexing in adjacent difference

senior-zero:
This PR is a preparation for addressing comments in the following [PR](https://github.com/NVIDIA/thrust/pull/1512).

### DIFF
--- a/cub/device/device_adjacent_difference.cuh
+++ b/cub/device/device_adjacent_difference.cuh
@@ -27,9 +27,10 @@
 
 #pragma once
 
-#include "../config.cuh"
-#include "../util_namespace.cuh"
-#include "dispatch/dispatch_adjacent_difference.cuh"
+#include <cub/config.cuh>
+#include <cub/detail/choose_offset.cuh>
+#include <cub/device/dispatch/dispatch_adjacent_difference.cuh>
+#include <cub/util_namespace.cuh>
 
 #include <thrust/detail/integer_traits.h>
 #include <thrust/detail/cstdint.h>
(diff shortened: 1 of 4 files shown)


cub/device/device_adjacent_difference.cuh
```diff
@@ -109,53 +109,35 @@ private:
...
-                     std::size_t num_items,
+                     NumItemsT num_items,
                      DifferenceOpT difference_op,
                      cudaStream_t stream,
                      bool debug_synchronous)
   {
-    const auto uint64_threshold = static_cast<std::size_t>(
-      THRUST_NS_QUALIFIER::detail::integer_traits<
-        THRUST_NS_QUALIFIER::detail::int32_t>::const_max);
+    static_assert(
```
allisonvacanti, line 117: Instead of reimplementing this logic for each Device algorithm, let's take the `ChooseOffsetT` helper that Andy added to `device_radix_sort.cuh`, move it to `cub/detail/choose_offset.cuh`, and reuse it here.

test/test_device_adjacent_difference.cu
```diff
@@ -491,17 +502,24 @@ void Test(std::size_t elements)
...
 }
 
 
-template <typename ValueT>
-void TestFancyIterators(std::size_t elements)
+template <typename ValueT,
+          typename NumItemsT>
+void TestFancyIterators(NumItemsT elements)
 {
+  if (elements == 0)
```
allisonvacanti, line 518: Is this needed? We should test that empty inputs are well behaved.

allisonvacanti (approved):
Let's reuse the `ChooseOffsetT` abstraction, but once that's done this is ready to test + merge.

cub/device/device_adjacent_difference.cuh
senior-zero: Good idea, thanks! I've made one more change in the extracted `ChooseOffset`, though. I've replaced `uint32_t` with `std::uint32_t`.

test/test_device_adjacent_difference.cu
senior-zero: It's needed because this particular test doesn't work in case of `elements == 0` (uses `elements - 1`). Empty inputs are tested in `Test` and `TestCopy`. Iterators doesn't affect the underlying logic, so it's safe to skip here.

Issue #478, opened 2022-05-11
Need a macro to disable bf16 support?


leofang:
After #306 CUB supports bf16 in some functions. Unfortunately, this means the bf16 headers (`cuda_bf16.h`, `cuda_bf16.hpp`) become a dependency of Thrust/CUB. Let me explain why it is unfortunate. 

Normally, this would be fine in the majority of cases, as CUDA headers are also needed at compile time. But in some cases, such as within a CUDA runtime docker (which contains no headers) + using Jitify to compile CUB kernels at runtime, it is a big issue due to the missing headers. 

Usually, the community approach is to bundle certain headers and redistribute them, so for example Thrust, CUB, and the fp16 headers (`cuda_fp16.h`, `cuda_fp16.hpp`) can be (and are being) repackaged in many projects. The unfortunate point is that the bf16 headers are *not* redistributable according to the CUDA EULA, and based on an internal conversation it'd remain the case for a long while. Therefore, it would be nice if CUB could offer a macro to disable the bf16 support.

PR #480, opened 2022-05-11
Add option to disable BF16 support

senior-zero:
This PR addresses the following [issue](https://github.com/NVIDIA/cub/issues/478). I considered eliminating of `cuda_bf16.h` header inclusion, but it might lead to issues when algorithm header relies on some different header:
```cpp
#include <cub/util_type.cuh> // omitted bf16 specializations based on `__CUDA_BF16_TYPES_EXIST__`
#include <cuda_bf16.h> // defined `__CUDA_BF16_TYPES_EXIST__`
#include <cub/device/device_radix_sort.cuh> // expects bf16 specializations in `util_type.h` 
```
It seems to be safer to provide the disable macro that @leofang's suggested.

### DIFF
--- a/cmake/CubHeaderTesting.cmake
+++ b/cmake/CubHeaderTesting.cmake
@@ -20,19 +20,27 @@ foreach (header IN LISTS headers)
   list(APPEND headertest_srcs "${headertest_src}")
 endforeach()
 
-foreach(cub_target IN LISTS CUB_TARGETS)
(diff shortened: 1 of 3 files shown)


allisonvacanti (approved):

allisonvacanti:
> Would this be included in a 1.x patch release, or it's 2.0 only?

It'll be in 2.0.

Issue #353, opened 2021-08-02
DeviceRadixSort fails when begin_bit = end_bit = 0 (for large inputs)


benbarsdell:
We had a painful issue in TensorFlow that turned out to be because we were passing `begin_bit = end_bit = 0` (all keys were zero in our case). CUB failed with "Invalid configuration error", and debugging was difficult because the failing kernel launch was not logged even with `debug_synchronous=true`.

Some isolated testing show that CUB succeeds and gives the correct answer for small inputs (i.e., single-block), but for large inputs it either produces the wrong value (CUDA <= 11.2) or returns "Invalid configuration error" (CUDA >= 11.3).

Here is a minimal reproducer (remove the ".txt" suffix):
[test_cub_bits_bug.cu.txt](https://github.com/NVIDIA/cub/files/6916640/test_cub_bits_bug.cu.txt)

It would be great if this could be fixed, and if logging could be added for all kernel launches (specifically in the Onesweep path).

senior-zero:
Hello, @benbarsdell! Thank you for reporting this. As stated in the documentation, key bits should be different:
> An optional bit subrange <tt>[begin_bit, end_bit)</tt> of differentiating key bits can be specified.

I think that's the reason there are no tests for this particular case:

```cuda
for (int begin_bit = 0; begin_bit <= 1; begin_bit++)
{
    // Iterate end bit
    for (int end_bit = begin_bit + 1;
```

As I understand, you expect something like cudaMemcpyAsync to be performed in this case. I think we could generalize API to this case at some point. Can you use some wrapper function until then?

benbarsdell:
Yes I've worked around it, so it's not blocking us.

I think it would be useful to generalize the API to support it. It's not actually clear to me that that docstring excludes this case, because it does not say that the subrange must be non-empty. We are calling it with `end_bit = Log2Ceiling(N)`, for integer keys in the range `[0, N)`, which results in `end_bit = 0` when `N = 1`.
There is also the fact that it already works in the single-block path.

allisonvacanti:
This seems reasonable. We should handle the case where `begin_bit == end_bit` to simplify generic usecases and add tests for this.

I opened #355 to track the missing log output.

PR #481, opened 2022-05-12
Fix begin_bit == end_bit == 0 for device-wide and segmented sort

canonizer:
Fix `begin_bit == end_bit == 0` for device-wide and segmented sort.

### DIFF
--- a/cub/device/dispatch/dispatch_radix_sort.cuh
+++ b/cub/device/dispatch/dispatch_radix_sort.cuh
@@ -912,12 +912,12 @@ struct DeviceRadixSortPolicy
     struct Policy800 : ChainedPolicy<800, Policy800, Policy700>
     {
         enum {
-            PRIMARY_RADIX_BITS      = (sizeof(KeyT) > 1) ? 7 : 5,
-            SINGLE_TILE_RADIX_BITS  = (sizeof(KeyT) > 1) ? 6 : 5,
-            SEGMENTED_RADIX_BITS    = (sizeof(KeyT) > 1) ? 6 : 5,
-            ONESWEEP = sizeof(KeyT) >= sizeof(uint32_t),
-            ONESWEEP_RADIX_BITS = 8,
-            OFFSET_64BIT            = sizeof(OffsetT) == 8,
+            PRIMARY_RADIX_BITS     = (sizeof(KeyT) > 1) ? 7 : 5,
+            SINGLE_TILE_RADIX_BITS = (sizeof(KeyT) > 1) ? 6 : 5,
+            SEGMENTED_RADIX_BITS   = (sizeof(KeyT) > 1) ? 6 : 5,
+            ONESWEEP               = sizeof(KeyT) >= sizeof(uint32_t),
+            ONESWEEP_RADIX_BITS    = 8,
+            OFFSET_64BIT           = sizeof(OffsetT) == 8,
         };
 
         // Histogram policy
@@ -1366,7 +1366,7 @@ struct DispatchRadixSort :
         ValueT* d_values_tmp2 = (ValueT*)allocations[3];
         AtomicOffsetT* d_ctrs = (AtomicOffsetT*)allocations[4];
 
-        do { 
+        do {
             // initialization
             if (CubDebug(error = cudaMemsetAsync(
                    d_ctrs, 0, num_portions * num_passes * sizeof(AtomicOffsetT), stream))) break;
@@ -1498,6 +1498,8 @@ struct DispatchRadixSort :
                     }
                 }
 
+                if (CubDebug(error)) break;
+                
                 // use the temporary buffers if no overwrite is allowed
                 if (!is_overwrite_okay && pass == 0)
                 {
@@ -1671,6 +1673,42 @@ struct DispatchRadixSort :
         return InvokeOnesweep<ActivePolicyT>();
     }
 
+    CUB_RUNTIME_FUNCTION __forceinline__
+    cudaError_t InvokeCopy()
+    {
+        // is_overwrite_okay == false here
+        // Return the number of temporary bytes if requested
+        if (d_temp_storage == nullptr)
+        {
+            temp_storage_bytes = 1;
+            return cudaSuccess;
+        }
+        
+        // Copy keys
+        cudaError_t error = cudaSuccess;
+        error = cudaMemcpyAsync(d_keys.Alternate(), d_keys.Current(), num_items * sizeof(KeyT),
+                                cudaMemcpyDefault, stream);
+        if (CubDebug(error))
+        {
+            return error;
+        }
+        d_keys.selector ^= 1;
+
+        // Copy values if necessary
+        if (!KEYS_ONLY)
+        {
+            error = cudaMemcpyAsync(d_values.Alternate(), d_values.Current(),
+                                    num_items * sizeof(ValueT), cudaMemcpyDefault, stream);
+            if (CubDebug(error))
+            {
+                return error;
+            }
+        }
+        d_values.selector ^= 1;
+
+        return error;
+    }
+
     /// Invocation
     template <typename ActivePolicyT>
     CUB_RUNTIME_FUNCTION __forceinline__
@@ -1679,15 +1717,20 @@ struct DispatchRadixSort :
         typedef typename DispatchRadixSort::MaxPolicy       MaxPolicyT;
         typedef typename ActivePolicyT::SingleTilePolicy    SingleTilePolicyT;
 
-        // Return if empty problem
-        if (num_items == 0)
+        // Return if empty problem, or if no bits to sort and double-buffering is used
+        if (num_items == 0 || (begin_bit == end_bit && is_overwrite_okay))
         {
-          if (d_temp_storage == nullptr)
-          {
-            temp_storage_bytes = 1;
-          }
+            if (d_temp_storage == nullptr)
+            {
+                temp_storage_bytes = 1;
+            }
+            return cudaSuccess;
+        }
 
-          return cudaSuccess;
+        // Check if simple copy suffices (is_overwrite_okay == false at this point)
+        if (begin_bit == end_bit)
+        {
+            return InvokeCopy();
         }
 
         // Force kernel code-generation in all compiler passes
@@ -2021,7 +2064,7 @@ struct DispatchSegmentedRadixSort :
             int radix_bits          = ActivePolicyT::SegmentedPolicy::RADIX_BITS;
             int alt_radix_bits      = ActivePolicyT::AltSegmentedPolicy::RADIX_BITS;
             int num_bits            = end_bit - begin_bit;
-            int num_passes          = (num_bits + radix_bits - 1) / radix_bits;
+            int num_passes          = CUB_MAX(DivideAndRoundUp(num_bits, radix_bits), 1);
             bool is_num_passes_odd  = num_passes & 1;
             int max_alt_passes      = (num_passes * radix_bits) - num_bits;
             int alt_end_bit         = CUB_MIN(end_bit, begin_bit + (max_alt_passes * alt_radix_bits));
@@ -2082,15 +2125,14 @@ struct DispatchSegmentedRadixSort :
     {
         typedef typename DispatchSegmentedRadixSort::MaxPolicy MaxPolicyT;
 
-        // Return if empty problem
-        if (num_items == 0)
+        // Return if empty problem, or if no bits to sort and double-buffering is used
+        if (num_items == 0 || (begin_bit == end_bit && is_overwrite_okay))
         {
-          if (d_temp_storage == nullptr)
-          {
-            temp_storage_bytes = 1;
-          }
-
-          return cudaSuccess;
+            if (d_temp_storage == nullptr)
+            {
+                temp_storage_bytes = 1;
+            }
+            return cudaSuccess;
         }
 
         // Force kernel code-generation in all compiler passes
--- a/cub/device/dispatch/dispatch_radix_sort.cuh
+++ b/cub/device/dispatch/dispatch_radix_sort.cuh
@@ -1498,7 +1498,10 @@ struct DispatchRadixSort :
                     }
                 }
 
-                if (CubDebug(error)) break;
+                if (error != cudaSuccess)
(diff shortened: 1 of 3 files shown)


canonizer:
This pull request should address https://github.com/NVIDIA/cub/issues/353.

senior-zero (changes requested):
@canonizer thank you for addressing this! I'm a bit concerned about the approach, though. I wonder if we could short-circuit in the `begin_bit == end_bit` case. For instance, if `is_overwrite_okay == true` we wouldn't do anything, since the double buffer would contain proper data already. Otherwise, we might just copy the data. I've written a simple benchmark below that might help understand the impact of this approach. For double buffer case, we have noop, which is definitely faster. Otherwise, memcpy is about 40% faster than actually sorting anything.   

```cpp
#include <cub/cub.cuh>

#include <thrust/device_vector.h>

#include <iostream>

void sort(
    std::uint8_t *d_temp_storage, std::size_t &temp_storage_bytes,
    int *d_keys_in, int *d_keys_out, 
    int num_items, 
    bool use_buffer, bool short_circuit) 
{
  const int begin_bit = 0;
  const int end_bit = begin_bit;
  cub::DoubleBuffer<int> d_keys(d_keys_in, d_keys_out);

  if (use_buffer) {
    if (short_circuit) {
      temp_storage_bytes = 1; // noop
    } else {
      cub::DeviceRadixSort::SortKeys(
          d_temp_storage, temp_storage_bytes, 
          d_keys, num_items, begin_bit, end_bit);
    }
  } else {
    if (short_circuit) {
      if (d_temp_storage == nullptr) {
        temp_storage_bytes = 1; 
      } else {
        cudaMemcpy(d_keys_out, d_keys_in, sizeof(int) * num_items, cudaMemcpyDeviceToDevice);
      }
    } else {
      cub::DeviceRadixSort::SortKeys(
          d_temp_storage, temp_storage_bytes, 
          d_keys_in, d_keys_out, num_items, begin_bit, end_bit);
    }
  }
}

int main()
{
  const int num_items = 128 * 1024 * 1024;
  thrust::device_vector<int> in(num_items);
  thrust::device_vector<int> out(num_items);

  int *d_keys_in = thrust::raw_pointer_cast(in.data());
  int *d_keys_out = thrust::raw_pointer_cast(out.data());

  std::uint8_t *d_temp_storage{};
  std::size_t temp_storage_bytes = 0;

  const bool use_buffer = false;
  const bool short_circuit = true;

  sort(d_temp_storage, temp_storage_bytes, d_keys_in, d_keys_out, num_items,
       use_buffer, short_circuit);

  thrust::device_vector<std::uint8_t> temp_storage(temp_storage_bytes);
  d_temp_storage = thrust::raw_pointer_cast(temp_storage.data());

  cudaEvent_t begin, end;
  cudaEventCreate(&begin);
  cudaEventCreate(&end);

  cudaEventRecord(begin);
  sort(d_temp_storage, temp_storage_bytes, d_keys_in, d_keys_out, num_items,
       use_buffer, short_circuit);
  cudaEventRecord(end);
  cudaEventSynchronize(end);

  float ms{};
  cudaEventElapsedTime(&ms, begin, end);

  std::cout << ms << "ms" << std::endl;


  cudaEventDestroy(end);
  cudaEventDestroy(begin);
}
```

I don't think this approach is applicable to segmented version. But I'd like to know your opinion on this for non-segmented API. Are there any downsides I'm missing?

canonizer:
@senior-zero @allisonvacanti I've added short-circuiting when `begin_bit == end_bit` and double buffers are passed as arguments. Could you take another look?

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1575,15 +1580,11 @@ struct DispatchRadixSort :
...
+        // Return if empty problem, or if no bits to sort and double-buffering is used
+        if (num_items == 0 || begin_bit == end_bit && is_overwrite_okay)
         {
-          if (d_temp_storage == nullptr)
-          {
-            temp_storage_bytes = 1;
-          }
-
-          return cudaSuccess;
+            if (d_temp_storage == nullptr) temp_storage_bytes = 1;
```
allisonvacanti, line 1586: Leave the explicit braces / linebreaks in the body here -- we're in the process of migrating code to use the safer style and want the body explicitly bounded.

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1925,15 +1926,11 @@ struct DispatchSegmentedRadixSort :
...
```
allisonvacanti, line 1932: Same as above.

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1575,15 +1580,11 @@ struct DispatchRadixSort :
         typedef typename DispatchRadixSort::MaxPolicy       MaxPolicyT;
         typedef typename ActivePolicyT::SingleTilePolicy    SingleTilePolicyT;
 
-        // Return if empty problem
-        if (num_items == 0)
+        // Return if empty problem, or if no bits to sort and double-buffering is used
+        if (num_items == 0 || begin_bit == end_bit && is_overwrite_okay)
```
allisonvacanti, line 1584: Explicit parens make intent clearer here:

```suggestion
        if (num_items == 0 || (begin_bit == end_bit && is_overwrite_okay))
```

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1925,15 +1926,11 @@ struct DispatchSegmentedRadixSort :
     {
         typedef typename DispatchSegmentedRadixSort::MaxPolicy MaxPolicyT;
 
-        // Return if empty problem
-        if (num_items == 0)
+        // Return if empty problem, or if no bits to sort and double-buffering is used
+        if (num_items == 0 || begin_bit == end_bit && is_overwrite_okay)
```
allisonvacanti, line 1930:
```suggestion
        if (num_items == 0 || (begin_bit == end_bit && is_overwrite_okay))
```

allisonvacanti (changes requested):
In @senior-zero's earlier comment, he suggested making the `is_overwrite_okay == false` case to just do a copy and skip the sorting altogether. Can we add that optimization?

cub/device/dispatch/dispatch_radix_sort.cuh
canonizer: Done.

cub/device/dispatch/dispatch_radix_sort.cuh
canonizer: Done.

cub/device/dispatch/dispatch_radix_sort.cuh
canonizer: Done.

cub/device/dispatch/dispatch_radix_sort.cuh
canonizer: Done.

canonizer:
Thanks for your comments!

@allisonvacanti I've addressed your comments.

@senior-zero I've added the copy shortcut if `begin_bit == end_bit`, and overwrite is not allowed, for the device-wide sort only. For the segmented sort, as you've mentioned, this approach won't work, and adding a separate kernel to perform segmented copy is definitely for a different pull request.

Could you take another look?

allisonvacanti:
@canonizer Can you rebase this on `main`? It looks like there are some conflicts.

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1451,6 +1451,8 @@ struct DispatchRadixSort :
                     }
                 }
 
+                if (CubDebug(error)) break;
```
senior-zero, line 1454: Is this check needed?

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1637,6 +1639,42 @@ struct DispatchRadixSort :
...
+        // Return the number of temporary bytes if requested
+        if (d_temp_storage == nullptr)
+        {
+            temp_storage_bytes = 1;
+            return cudaSuccess;
+        }
+        
+        // Copy keys
+        cudaError_t error = cudaSuccess;
+        error = cudaMemcpyAsync(d_keys.Alternate(), d_keys.Current(), num_items * sizeof(KeyT),
```
senior-zero, line 1655: We have to synchronize in the case of `debug_synchronous`
```
if (debug_synchronous)
{
  if (CubDebug(error = SyncStream(stream))
  {
    // ..
```

cub/device/dispatch/dispatch_radix_sort.cuh
canonizer: The check is necessary to break out of the loop in case of error. An extra `CubDebug()` is not, so I removed it.

cub/device/dispatch/dispatch_radix_sort.cuh
canonizer: Done (using `DebugSyncStream()`).

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1637,6 +1639,42 @@ struct DispatchRadixSort :
...
+        if (CubDebug(error))
+        {
+            return error;
+        }
+        d_keys.selector ^= 1;
+
+        // Copy values if necessary
+        if (!KEYS_ONLY)
+        {
+            error = cudaMemcpyAsync(d_values.Alternate(), d_values.Current(),
```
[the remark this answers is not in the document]
canonizer: Done (using `DebugSyncStream()`).

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1637,6 +1639,42 @@ struct DispatchRadixSort :
...
+        // is_overwrite_okay == false here
+        // Return the number of temporary bytes if requested
+        if (d_temp_storage == nullptr)
+        {
+            temp_storage_bytes = 1;
+            return cudaSuccess;
+        }
+        
+        // Copy keys
+        cudaError_t error = cudaSuccess;
```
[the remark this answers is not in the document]
canonizer: Done.

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1637,6 +1639,42 @@ struct DispatchRadixSort :
...
+        error = cudaMemcpyAsync(d_keys.Alternate(), d_keys.Current(), num_items * sizeof(KeyT),
+                                cudaMemcpyDefault, stream);
```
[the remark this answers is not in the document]
canonizer: Replaced with `cudaMempcyDeviceToDevice`.

canonizer:
@allisonvacanti @senior-zero I've synced with the latest `main` branch and addressed all review comments. Could you take another look?

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1747,7 +1747,10 @@ struct DispatchRadixSort :
         }
 
         // Check if simple copy suffices (is_overwrite_okay == false at this point)
-        if (begin_bit == end_bit)
+        cudaError_t error = cudaSuccess;
+        bool has_uva = false;
+        if ((error = HasUVA(has_uva)) != cudaSuccess) return error;
+        if (begin_bit == end_bit & has_uva)
```
senior-zero, line 1753: This check is not synchronized with the temp storage size estimation above. On systems without uva it'll lead to invalid memory accesses. 

Also, property queries are not free. Can we invoke `HasUVA` only when `begin_bit == end_bit`?

cub/util_device.cuh
```diff
@@ -125,6 +125,23 @@ CUB_RUNTIME_FUNCTION inline int CurrentDevice()
     return device;
 }
 
+/** \brief Gets whether the current device supports unified addressing */
+CUB_RUNTIME_FUNCTION cudaError_t HasUVA(bool& has_uva)
```
senior-zero, line 129: Can you put this function in the `detail` namespace for now?

senior-zero (changes requested):

cub/util_device.cuh
canonizer: Done.

cub/device/dispatch/dispatch_radix_sort.cuh
canonizer: Done (for invoking `HasUVA()` only when `begin_bit == end_bit`).

Regarding temporary storage estimate, the one above only applies to the cases where no GPU work at all is performed. If a single copy is required, a separate estimate is provided by `InvokeCopy()` (see the function definition for the actual code).

canonizer:
@senior-zero I've addressed your comments. Could you take another look?

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -1498,6 +1498,11 @@ struct DispatchRadixSort :
                     }
                 }
 
+                if (error != cudaSuccess)
```
senior-zero, line 1501: Everything that sets `error` before this place, also checks it, right? I fail to see why this check is needed.

canonizer: This check is needed to break out of the `for (int current_bit = ...)` loop in case of error.

senior-zero: Ah, got it, thanks!

senior-zero (approved):
Thank you for this optimization! I'll start testing now.

cub/util_device.cuh
```diff
@@ -549,6 +549,23 @@ CUB_RUNTIME_FUNCTION inline cudaError_t DebugSyncStream(cudaStream_t stream)
 #endif
 }
 
+/** \brief Gets whether the current device supports unified addressing */
+CUB_RUNTIME_FUNCTION cudaError_t HasUVA(bool& has_uva)
```
senior-zero, line 553:
```suggestion
CUB_RUNTIME_FUNCTION inline cudaError_t HasUVA(bool& has_uva)
```

PR #483, opened 2022-05-13
Allow device lambdas in device adjacent difference

senior-zero:
This PR allows `__device__` extended lambdas usage in the device adjacent difference (nvbug 3641330). The issue was caused by `invoke_result_t` usage on `__device__` extended lambda on the host side.

### DIFF
--- a/cub/device/dispatch/dispatch_adjacent_difference.cuh
+++ b/cub/device/dispatch/dispatch_adjacent_difference.cuh
@@ -64,7 +64,6 @@ template <typename ChainedPolicyT,
           typename DifferenceOpT,
           typename OffsetT,
           typename InputT,
-          typename OutputT,
           bool MayAlias,
           bool ReadLeft>
 void __global__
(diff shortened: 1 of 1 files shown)


senior-zero:
cc @zasdfgbnm

allisonvacanti (approved):
In general, we should be advising users to not use `__device__`-only extended lambdas, as they are extremely fragile footguns for a template library that relies so heavily on type deduction.

Since this case has an easy WAR to fix a regression I'm ok with the change.

@zasdfgbnm PyTorch should consider moving away from `__device__` lambdas to avoid these sorts of issues in the future. Using `__host__ __device__` lambdas is ok. If the function must be device-only, a function object should be used. This is due to a limitation of NVCC's implementation of lambdas that we cannot always WAR/fix on our end.

This issue is discussed in NVIDIA/thrust#779 and @jrhemstad is looking into improving the diagnostics in NVIDIA/thrust#1688.

zasdfgbnm:
Thanks for the quick fix! I think one reason why we don't like device host lambda is that for some reason it has a bigger binary size, and PyTorch has been working very hard to reduce its binary size. But functors sound like a good thing to use.

In the future, if a similar thing happens, our CI can detect it quickly, and I will confirm with you whether there is an easy way to fix it in cub/thrust. If unfortunately not, then we will update PyTorch to use functors.

allisonvacanti:
Thanks for understanding -- it's an unfortunate situation, and I'm looking forward to the new diagnostics that will help catch these sooner!

> In the future, if a similar thing happens, our CI can detect it quickly

It's great that ya'll are testing against `HEAD` -- I was happy to see that this was detected so soon after merging :)

allisonvacanti:
Backported to `1.17.X` and tagged 1.17.0-rc1:

https://github.com/NVIDIA/cub/commit/28de483b4e76957e9a860b9792c6ad8809ac1ecd

PR #486, opened 2022-05-18
Update CDP support macros for if-target compatibility

allisonvacanti:
Goes with NVIDIA/thrust#1661. Updates the CDP/CUDART macros to be consistent across host/device passes and thus compatible with NV_IF_TARGET.

All device algorithm tests with existing CDP tests have been updated to exercise this functionality. Some device algorithm tests do not have CDP tests and will need to be updated in a future PR.

# Pre-written Release Notes

## Breaking Changes

- NVIDIA/cub#486: CUB’s CUDA Runtime support macros have been updated to support `NV_IF_TARGET`. They are now defined consistently across all host/device compilation passes. This should not affect most usages of these macros, but may require changes for some edge cases.
    - `CUB_RUNTIME_FUNCTION`: Execution space annotations for functions that invoke CUDA Runtime APIs.
        - Old behavior:
            - RDC enabled: Defined to `__host__ __device__`
            - RDC not enabled:
                - NVCC host pass: Defined to `__host__ __device__`
                - NVCC device pass: Defined to `__host__`
        - New behavior:
            - RDC enabled: Defined to `__host__ __device__`
            - RDC not enabled: Defined to `__host__`
    - `CUB_RUNTIME_ENABLED`: No change in behavior, but no longer used in CUB. Provided for legacy support only. Legacy behavior:
        - RDC enabled: Macro is defined.
        - RDC not enabled:
            - NVCC host pass: Macro is defined.
            - NVCC device pass: Macro is not defined.
    - `CUB_RDC_ENABLED`: New macro, may be combined with `NV_IF_TARGET` to replace most usages of `CUB_RUNTIME_ENABLED`. Behavior:
        - RDC enabled: Macro is defined.
        - RDC not enabled: Macro is not defined.

## Other Enhancements

- NVIDIA/cub#486: Synchronizations triggered by the `debug_synchronous` flag in the device algorithm layer now print a log message.


cub/device/dispatch/dispatch_segmented_sort.cuh
```diff
@@ -1653,7 +1658,7 @@ private:
             return error;
         }
 
-        if (CubDebug(error = SyncStream(stream)))
+        if (CubDebug(error = DebugSyncStream(stream)))
```
senior-zero, line 1661: This sync is **not** meant as debug one. It's required for correctness, which contradicts the documentation of the `DebugSyncStream`. While reading the rest of the PR, I was thinking that `DebugSyncStream` usage is a bit bloated and unsafe. It's impossible to guarantee that the `DebugSyncStream` is called inside `if(debug)`. In order to guarantee this, can we change the signature to something like `DebugSyncStream(stream, debug)` and hide the branch inside? This would address mentioned concerns.

test/test_device_radix_sort.cu
```diff
@@ -490,104 +495,214 @@ cudaError_t Dispatch(
...
+// Specializations of Dispatch that translate the CDP backend to the appropriate
+// CUB backend, and uses the CUB backend to launch the CDP kernel.
+#define DEFINE_CDP_DISPATCHER(CdpBackend, CubBackend)                          \
+  template <int IsDescending,                                                  \
+            typename KeyT,                                                     \
+            typename ValueT,                                                   \
+            typename BeginOffsetIteratorT,                                     \
+            typename EndOffsetIteratorT,                                       \
+            typename NumItemsT>                                                \
+  CUB_RUNTIME_FUNCTION cudaError_t Dispatch(                                   \
```
senior-zero, line 659: Does this `Dispatch` function and `LaunchCDPKernel` has to be annotated as `CUB_RUNTIME_FUNCTION`? Could they always be `__host__` ones? This would allow us to remove `IF_TARGET` from these functions, right?

test/test_device_reduce.cu
```diff
@@ -521,91 +544,188 @@ cudaError_t Dispatch(
...
-    ReductionOpT        reduction_op,
-    cudaStream_t        stream,
-    bool                debug_synchronous)
+template <int CubBackend,
+          typename InputIteratorT,
+          typename OutputIteratorT,
+          typename BeginOffsetIteratorT,
+          typename EndOffsetIteratorT,
+          typename ReductionOpT>
+CUB_RUNTIME_FUNCTION cudaError_t
```
senior-zero, line 604: Same question as above. If there's something about marking `LaunchCDPKernel` as `CUB_RUNTIME_FUNCTION` that I'm missing, please ignore this. Otherwise, fix other places below as well.

senior-zero (approved):
This is a significant improvement and a huge work. Thank you for your efforts! There are but a few minor comments below.

cub/device/dispatch/dispatch_segmented_sort.cuh
allisonvacanti: Good suggestion. Switched this to a regular `SyncStream` and refactored `DebugSyncStream` to take a `debug_synchronous` argument.

test/test_device_radix_sort.cu
allisonvacanti: Oooh, nice suggestion, I hadn't considered that and just matched the signature of the other `Dispatch` overloads. This cleans things up a lot! 👍

PR #488, opened 2022-05-20
Wrap `AgentScanByKey::TempStorage` in `cub::Uninitialized`.

allisonvacanti:
This fixes invalid memory accesses in two Thrust examples (`scan_by_key` and `summed_area_table`) on optimized nvc++ builds.


dkolsen-pgi:
I don't know why this makes a difference.  But it doesn't look wrong.

dkolsen-pgi (approved):

PR #490, opened 2022-05-23
In-place guarantees for scan algorithms

senior-zero:
This PR addresses multiple issues of in-place execution for scan algorithms:

Scan:
- `LOAD_LDG` is removed from tuning policies to guarantee in-place execution. Affected architectures are Kepler and Maxwell. Benchmarks showed no difference in performance on those platforms. 
- In-place execution guarantees are highlighted in the documentation of the existing API
- In-place API is provided
- In-place tests are fixed. The code used to compare `decltype(*d_in)` with `OutputT` which always led to comparing `T&` with `T`. Therefore, we've never tested in-place execution. 

Scan by key:
- `LOAD_LDG` is removed from tuning policies to guarantee in-place execution. 
- To preserve thrust documentation guarantees, implementation is changed to allow aliasing of keys/results. This change affects performance on small to moderate problem sizes (up to 6% slowdown). 
- In-place execution guarantees are highlighted in the documentation of the existing API
- Tests of in/out values aliasing are fixed. The code used to compare `decltype(*d_values_in)` with `OutputT` which always led to comparing `T&` with `T`. Therefore, we've never tested in-place execution. 
- Tests of keys/values aliasing are added.

### DIFF
--- a/cub/device/device_scan.cuh
+++ b/cub/device/device_scan.cuh
@@ -1,7 +1,6 @@
-
 /******************************************************************************
  * Copyright (c) 2011, Duane Merrill.  All rights reserved.
- * Copyright (c) 2011-2018, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 2011-2022, NVIDIA CORPORATION.  All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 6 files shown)


cub/device/device_scan.cuh
```diff
@@ -28,844 +27,1974 @@
...
+   * @brief Computes a device-wide exclusive prefix sum. The value of `0` is 
+   *        applied as the initial value, and is assigned to `*d_out`.
+   *
+   * @par
+   * - Supports non-commutative sum operators.
+   * - Results are not deterministic for pseudo-associative operators (e.g.,
+   *   addition of floating-point types). Results for pseudo-associative
+   *   operators may vary from run to run. Additional details can be found in
+   *   the [decoupled look-back] description.
+   * - When the input and output sequences are the same, the scan is
```
allisonvacanti, line 109: This point seems redundant with the following one.

cub/device/device_scan.cuh
```diff
@@ -28,844 +27,1974 @@
...
+   * @brief Computes a device-wide exclusive prefix sum in-place. The value of 
+   *        `0` is applied as the initial value, and is assigned to `*d_data`.
+   *
+   * @par
+   * - Supports non-commutative sum operators.
+   * - Results are not deterministic for pseudo-associative operators (e.g.,
+   *   addition of floating-point types). Results for pseudo-associative
+   *   operators may vary from run to run. Additional details can be found in
+   *   the [decoupled look-back] description.
+   * - When the input and output sequences are the same, the scan is
```
allisonvacanti, line 229: This bit should be removed from the in-place overload.

allisonvacanti (approved):
Some minor issues in doc strings, but otherwise LGTM.

cub/device/device_scan.cuh
senior-zero: Thanks for noticing this!

PR #500, opened 2022-05-31
In-place guarantees for reduce

senior-zero:
Reduce consists of two kernels. The first one reads input data and accumulates partial sums in a temporary storage. The second one reads the temporary storage and writes the final result. Therefore, it's safe to alias input and output arrays.
Nonetheless, allowing in-place execution would limit our abilities to optimize algorithm later. Since aliasing doesn't provide significant memory saving, I'd rather not allow it. 

Regarding `ByKey` variant, it relies on decoupled look back, so it should be safe to alias in/out data as long as value types for input/output iterators match exactly. The only limiting factor is `LOAD_LDG` which makes aliasing in this case an UB. In-place execution in this case would provide significant memory savings. If there's a request, I suggest we add an overload that would allow in-place execution. 

Regarding `Segmented` version, one block is assigned per segment. Results are written without synchronization between blocks, therefore, any aliasing with output array would introduce a data race.

### DIFF
--- a/cub/device/device_segmented_reduce.cuh
+++ b/cub/device/device_segmented_reduce.cuh
@@ -1,7 +1,6 @@
-
 /******************************************************************************
  * Copyright (c) 2011, Duane Merrill.  All rights reserved.
- * Copyright (c) 2011-2018, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 2011-2022, NVIDIA CORPORATION.  All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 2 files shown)


cub/device/device_segmented_reduce.cuh
```diff
@@ -28,606 +27,1018 @@
...
+    using OutputValueT = typename OutputTupleT::Value;
+
+    // Wrapped input iterator to produce index-value <OffsetT, InputT> tuples
+    using ArgIndexInputIteratorT =
+      ArgIndexInputIterator<InputIteratorT, OffsetT, OutputValueT>;
+
+    ArgIndexInputIteratorT d_indexed_in(d_in);
+
+    // Initial value
+    OutputTupleT initial_value(1, Traits<InputValueT>::Lowest()); // replace
```
allisonvacanti, line 1017: This trailing comment should be reformatted, clang-format is being weird 😄

cub/device/device_segmented_reduce.cuh
```diff
@@ -28,606 +27,1018 @@
...
+                          debug_synchronous);
+  }
+
+  /**
+   * @brief Finds the first device-wide maximum in each segment using the 
+   *        greater-than ('>') operator, also returning the in-segment index of 
+   *        that item
+   *
+   * @par
+   * - The output value type of `d_out` is cub::KeyValuePair `<int, T>`
```
allisonvacanti, line 873:
```suggestion
   * - The output value type of `d_out` is `cub::KeyValuePair<int, T>` 
```

allisonvacanti (approved):
LGTM, just a couple of style nits.

PR #501, opened 2022-05-31
In-place guarantees for rle

senior-zero:
`Encode` doesn't allow aliasing for the same reason as reduce by key. `NonTrivialRuns` reads out-of-tile elements without synchronization. Aliasing in `NonTrivialRuns` would introduce data race.

### DIFF
--- a/cub/device/device_run_length_encode.cuh
+++ b/cub/device/device_run_length_encode.cuh
@@ -85,6 +85,12 @@ struct DeviceRunLengthEncode
    * - The total number of runs encountered is written to `d_num_runs_out`.
    * - The `==` equality operator is used to determine whether values are 
    *   equivalent
+   * - In-place operations are not supported. There must be no overlap between
+   *   any of the provided ranges:
+   *   - `[d_unique_out, d_unique_out + *d_num_runs_out)`
+   *   - `[d_counts_out, d_counts_out + *d_num_runs_out)`
+   *   - `[d_num_runs_out, d_num_runs_out + 1)`
+   *   - `[d_in, d_in + num_items)`
    * - @devicestorage
    *
    * @par Performance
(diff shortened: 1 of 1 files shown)


cub/device/device_run_length_encode.cuh
```diff
@@ -28,245 +27,366 @@
...
+   * @brief Computes a run-length encoding of the sequence \p d_in.
+   *
+   * @par
+   * - For the *i*<sup>th</sup> run encountered, the first key of the run and 
+   *   its length are written to `d_unique_out[i]` and `d_counts_out[i]`,
+   *   respectively.
+   * - The total number of runs encountered is written to `d_num_runs_out`.
+   * - The `==` equality operator is used to determine whether values are 
+   *   equivalent
+   * - Let `out` represent a set of following ranges
```
allisonvacanti, line 88: This seems equivalent to just stating that there shall be no overlap between any of the 4 listed ranges, with no need to group the first three together as `out`. Is there some nuance I'm missing here?

cub/device/device_run_length_encode.cuh
```diff
@@ -28,245 +27,366 @@
...
+   *        (of `length > 1`) of same-valued keys in the sequence `d_in`.
+   *
+   * @par
+   * - For the *i*<sup>th</sup> non-trivial run, the run's starting offset and 
+   *   its length are written to `d_offsets_out[i]` and `d_lengths_out[i]`, 
+   *   respectively.
+   * - The total number of runs encountered is written to `d_num_runs_out`.
+   * - The `==` equality operator is used to determine whether values are 
+   *   equivalent
+   * - Let `out` represent a set of following ranges
```
allisonvacanti, line 257: Same note as earlier.

cub/device/device_run_length_encode.cuh
senior-zero: There's no technical nuance you are missing. This phrasing just represented my mental model of these iterators :smile: The suggested phrasing might be easier to read, so I've updated the docs, thanks!

PR #503, opened 2022-06-03
In place select

senior-zero:
Select algorithms are based on the decoupled look back approach. Therefore, subsequent thread blocks are guaranteed to write data strictly after input data was read. Unlike partition family, select writes data only from one side of the array. It's safe to have `in` iterator equal to the `out` one. The only issue is `LOAD_LDG` that is used by default. `LOAD_LDG` replacement with `LOAD_CA` leads to 50% slowdown on Kepler and about 30% slowdown on Maxwell. To avoid performance regression on these architectures I've forbidden in-place execution and left `LOAD_LDG`. Since it'd be unfortunate to loose in-place option, I introduced in-place overload that takes exactly one argument. This also addresses the following [issue](https://github.com/NVIDIA/cub/issues/502).

The unique subset of algorithms reads data outside of thread block tile. This leads to data races. It's possible to introduce in-place version but it'd require more work (caching pre-tile data in temporary storage).

### DIFF
--- a/cub/device/device_select.cuh
+++ b/cub/device/device_select.cuh
@@ -90,6 +90,9 @@ struct DeviceSelect
    *   `char`, `int`, etc.).
    * - Copies of the selected items are compacted into `d_out` and maintain 
    *   their original relative ordering.
+   * - The range `[d_out, d_out + *d_num_selected_out)` shall not overlap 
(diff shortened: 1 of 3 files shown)


allisonvacanti (approved):

allisonvacanti:
LGTM -- run this through gpuCI to validate the new tests before merging.

Issue #428, opened 2022-01-28
Use P2322R6 to determine intermediate types for relevant algorithms


codecircuit:
The `reduction_op` is called with a type derived from the output iterator.  Example:

```cpp
#include <cub/cub.cuh>
#include <iostream>

struct CustomMin {
	template <typename T>
	__device__ __host__ __forceinline__ T operator()(const T& a,
	                                                 const T& b) const {
		return (b < a) ? b : a;
	}
};

struct result_t {
	result_t() = default;
	__device__ __host__ result_t(int i) : data(i) {
	}

	double data = 0;
};

inline std::ostream& operator<<(std::ostream& os, const result_t& r) {
	os << r.data;
	return os;
}

int main() {
	int* d_offsets;
	int* d_in;
	using OT = result_t;
	//using OT = int; // compiles
	OT* d_out;
	CustomMin min_op;

	int initial_value = INT_MAX;

	const int num_segments = 3;
	const int num_values = 7;

	cudaMallocManaged(&d_offsets, sizeof(int) * (num_segments + 1));
	cudaMallocManaged(&d_in, sizeof(int) * num_values);
	cudaMallocManaged(&d_out, sizeof(OT) * num_segments);

	d_offsets[0] = 0;
	d_offsets[1] = 3;
	d_offsets[2] = 3;
	d_offsets[3] = 7;

	d_in[0] = 8;
	d_in[1] = 6;
	d_in[2] = 7;
	d_in[3] = 5;
	d_in[4] = 3;
	d_in[5] = 0;
	d_in[6] = 9;

	void* d_temp_storage = NULL;
	size_t temp_storage_bytes = 0;

	cub::DeviceSegmentedReduce::Reduce(d_temp_storage,
	                                   temp_storage_bytes,
	                                   d_in,
	                                   d_out,
	                                   num_segments,
	                                   d_offsets,
	                                   d_offsets + 1,
	                                   min_op,
	                                   initial_value);

	cudaMalloc(&d_temp_storage, temp_storage_bytes);

	cub::DeviceSegmentedReduce::Reduce(d_temp_storage,
	                                   temp_storage_bytes,
	                                   d_in,
	                                   d_out,
	                                   num_segments,
	                                   d_offsets,
	                                   d_offsets + 1,
	                                   min_op,
	                                   initial_value);
	cudaDeviceSynchronize();

	for (int i = 0; i < num_segments; ++i) {
[...]

PR #509, opened 2022-06-13
P2322R6 accumulator types for reduce

senior-zero:
This PR addresses the following [issue](https://github.com/NVIDIA/cub/issues/428). I also found that we use copy assignment operator on uninitialized memory, which might lead to issues for not primitive types.

### DIFF
--- a/cub/agent/agent_reduce.cuh
+++ b/cub/agent/agent_reduce.cuh
@@ -1,6 +1,6 @@
 /******************************************************************************
  * Copyright (c) 2011, Duane Merrill.  All rights reserved.
- * Copyright (c) 2011-2018, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 2011-2022, NVIDIA CORPORATION.  All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
@@ -13,10 +13,10 @@
  *       names of its contributors may be used to endorse or promote products
  *       derived from this software without specific prior written permission.
  *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
- * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
- * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
- * DISCLAIMED. IN NO EVENT SHALL NVIDIA CORPORATION BE LIABLE FOR ANY
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" 
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE 
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+ * ARE DISCLAIMED. IN NO EVENT SHALL NVIDIA CORPORATION BE LIABLE FOR ANY
  * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
  * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
  * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
@@ -27,25 +27,24 @@
  ******************************************************************************/
 
 /**
- * \file
- * cub::AgentReduce implements a stateful abstraction of CUDA thread blocks for participating in device-wide reduction .
+ * @file cub::AgentReduce implements a stateful abstraction of CUDA thread 
+ *       blocks for participating in device-wide reduction.
  */
 
 #pragma once
 
 #include <iterator>
 
-#include "../block/block_load.cuh"
-#include "../block/block_reduce.cuh"
-#include "../grid/grid_mapping.cuh"
-#include "../grid/grid_even_share.cuh"
-#include "../config.cuh"
-#include "../util_type.cuh"
-#include "../iterator/cache_modified_input_iterator.cuh"
-
+#include <cub/block/block_load.cuh>
+#include <cub/block/block_reduce.cuh>
+#include <cub/config.cuh>
+#include <cub/detail/type_traits.cuh>
+#include <cub/grid/grid_even_share.cuh>
+#include <cub/grid/grid_mapping.cuh>
+#include <cub/iterator/cache_modified_input_iterator.cuh>
+#include <cub/util_type.cuh>
 
 CUB_NAMESPACE_BEGIN
-
 
 /******************************************************************************
  * Tuning policy types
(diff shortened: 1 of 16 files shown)


senior-zero:
This PR should be merged after libcu++ supports device lambdas in invoke result: https://github.com/NVIDIA/libcudacxx/pull/284

cub/device/dispatch/dispatch_reduce.cuh
```diff
@@ -55,788 +54,1247 @@ CUB_NAMESPACE_BEGIN
...
-    SelectedPolicy
+  typename InputIteratorT,
+  typename OutputIteratorT,
+  typename OffsetT,
+  typename ReductionOpT,
+  typename InitT = 
+    cub::detail::non_void_value_t<
+      OutputIteratorT, 
+      cub::detail::value_t<InputIteratorT>>,
+  typename AccumT =
```
allisonvacanti, line 483: Can you add the `release: breaking change` label and put a description of these changes to the Dispatch interface in the PR description? That way I'll make sure to call these changes out in the release notes.

Same goes for all of the accumulator-type changes in behavior -- I check that label when I'm building relnotes from the list of PRs.

allisonvacanti (approved):
This will need a couple of things before merging:

- Rebase for `debug_synchronous`/`DebugSyncStream`/`CDP` changes
- Add summary of breaking changes to PR description.

cub/agent/agent_reduce.cuh
```diff
@@ -27,357 +27,418 @@
...
+  // Utility
+  //---------------------------------------------------------------------
+
+  // Whether or not the input is aligned with the vector type (specialized for
+  // types we can vectorize)
+  template <typename Iterator>
+  static __device__ __forceinline__ bool
+  IsAligned(Iterator d_in, Int2Type<true> /*can_vectorize*/)
+  {
+    return (size_t(d_in) & (sizeof(VectorT) - 1)) == 0;
```
miscco, line 194: Shouldnt that be a property of the type `Iterator` and not the individual pointer `d_in`?

cub/agent/agent_reduce.cuh
```diff
@@ -27,357 +27,418 @@
...
+    // Load items in striped fashion
+    LoadDirectStriped<BLOCK_THREADS>(threadIdx.x,
+                                     d_wrapped_in + block_offset,
+                                     items);
+
+    // Reduce items within each thread stripe
+    thread_aggregate =
+      (IS_FIRST_TILE)
+        ? internal::ThreadReduce(items, reduction_op)
+        : internal::ThreadReduce(items, reduction_op, thread_aggregate);
```
miscco, line 254: My knowledge here is not yet up to date, so please excuse if this is stupid.

Why do we treat the first tile specially? From my understanding it should suffice to properly initialize `thread_aggregate` and treat all tiles equally

cub/agent/agent_reduce.cuh
```diff
@@ -27,357 +27,418 @@
...
+      , reduction_op(reduction_op)
+  {}
+
+  //---------------------------------------------------------------------
+  // Tile consumption
+  //---------------------------------------------------------------------
+
+  /**
+   * @brief Consume a full tile of input (non-vectorized)
+   * @param block_offset The offset the tile to consume
```
miscco, line 231: We are missing the `thread_aggregate` param or is that conventionally skipped?

cub/agent/agent_reduce.cuh
```diff
@@ -27,357 +27,418 @@
...
-
-    // Whether or not the input is aligned with the vector type (specialized for types we can vectorize)
-    template <typename Iterator>
-    static __device__ __forceinline__ bool IsAligned(
-        Iterator        d_in,
-        Int2Type<true>  /*can_vectorize*/)
-    {
-        return (size_t(d_in) & (sizeof(VectorT) - 1)) == 0;
+      vec_items[i] = d_vec_in[BLOCK_THREADS * i];
     }
```
miscco, line 290: No change requested: That seems like a really common operation. Should we have a helper function for it?

cub/device/dispatch/dispatch_reduce.cuh
```diff
@@ -55,898 +55,1291 @@ CUB_NAMESPACE_BEGIN
...
-    CUB_RUNTIME_FUNCTION __forceinline__
-    cudaError_t InvokeSingleTile(
-        SingleTileKernelT       single_tile_kernel)     ///< [in] Kernel function pointer to parameterization of cub::DeviceReduceSingleTileKernel
+      // Return if the caller is simply requesting the size of the storage
+      // allocation
+      if (d_temp_storage == NULL)
+      {
+        temp_storage_bytes = 1;
+        break;
+      }
```
miscco, line 604: This seems to be an invariant, so I would prefer if we could move it to the top of the function and turn the `break` into an explicit `return cudaSuccess;`

cub/device/dispatch/dispatch_reduce.cuh
```diff
@@ -55,898 +55,1291 @@ CUB_NAMESPACE_BEGIN
...
+
+      // Invoke single_reduce_sweep_kernel
+      THRUST_NS_QUALIFIER::cuda_cub::launcher::triple_chevron(
+        1, ActivePolicyT::SingleTilePolicy::BLOCK_THREADS, 0, stream)
+        .doit(single_tile_kernel, d_in, d_out, num_items, reduction_op, init);
+
+      // Check for failure to launch
+      if (CubDebug(error = cudaPeekAtLastError()))
+      {
+        break;
```
miscco, line 623: I am not a big fan of those break into return blocks. Could we simply do a `return error;`?

cub/device/dispatch/dispatch_reduce.cuh
```diff
@@ -55,898 +55,1291 @@ CUB_NAMESPACE_BEGIN
...
+      }
+
+      // Sync the stream if specified to flush runtime errors
+      error = detail::DebugSyncStream(stream);
+      if (CubDebug(error))
```
miscco, line 628: Above you pulled the assignment into the if. I believe we should settle on one style

cub/device/dispatch/dispatch_reduce.cuh
```diff
@@ -55,898 +55,1291 @@ CUB_NAMESPACE_BEGIN
...
-    template <typename ActivePolicyT>
-    CUB_RUNTIME_FUNCTION __forceinline__
-    cudaError_t Invoke()
+      // Return if the caller is simply requesting the size of the storage
+      // allocation
+      if (d_temp_storage == NULL)
+      {
+        temp_storage_bytes = 1;
+        return cudaSuccess;
+      }
```
miscco, line 1140: Ditto move in front of the loop

cub/thread/thread_operators.cuh
```diff
@@ -37,308 +37,324 @@
...
-    __host__ __device__ __forceinline__ bool operator()(const T &a, const T &b) const
-    {
-        return a == b;
-    }
+  /// Boolean equality operator, returns `t == u`
+  template <typename T, typename U>
+  __host__ __device__ __forceinline__ bool operator()(T &&t, U &&u) const
+  {
+    return cuda::std::forward<T>(t) == cuda::std::forward<U>(u);
+  }
```
miscco, line 62: Please tell me that there is nobody out there that specializes comparison operators depending on value categories.

Can we simply bind to a `const T&` and drop the forward?

Applies throughout

miscco (commented):
I am not sure if I am yet qualified to really approve it but I can complain

cub/agent/agent_reduce.cuh
senior-zero: I think the following code can illustrate what's going on here:
```cpp
#include <stdint.h>
#include <iostream>

template <class VectorT, class IteratorT>
bool is_aligned(IteratorT iter) {
  // can_vectorize = ... && is_pointer;
  return (size_t(iter) & (sizeof(VectorT) - 1)) == 0;
}

template <class VectorT, class IteratorT>
void foo(IteratorT iter) {
  std::cout << is_aligned<VectorT>(iter) << std::endl;
}

int main () {
  char data[32];
  foo<uint64_t>(data);
  foo<uint64_t>(data + 1);
}
```

How would you add property to the pointer? Are you suggesting wrapping pointers into a type that would perform `IsAligned` check in the constructor and provide a property to query?

cub/agent/agent_reduce.cuh
senior-zero: That's what the first overload is doing. Do you think that having:
```cpp
if (IS_FIRST_TILE) { thread_aggregate = input[0]; }
```
before the `ThreadReduce` invokation is better in any way?

cub/agent/agent_reduce.cuh
senior-zero: Good catch. I guess that only input parameters are documented here. Before refactoring, this was undocumented as well. I hope that we'll be able to rewrite this to return `thread_aggreate` instead of taking it as a parameter.

cub/agent/agent_reduce.cuh
senior-zero: I think we should, once this is used in more than one place.

cub/device/dispatch/dispatch_reduce.cuh
senior-zero: I'd prefer this as well. I think it might be a good idea to have a cleanup PR that removes all `do{} while(0);` from the code.

cub/device/dispatch/dispatch_reduce.cuh
senior-zero: I'd prefer consistency and small PRs in this case. We can have a cleanup PR later that would remove all `do{} while(0);` from the code.

cub/device/dispatch/dispatch_reduce.cuh
senior-zero: I haven't pulled the assignment into the if.

cub/device/dispatch/dispatch_reduce.cuh
senior-zero: There's no loop :smile: See my comments to the similar observations above.

cub/thread/thread_operators.cuh
senior-zero: What's the advantage in this case?

miscco: Better throughput, we have 2 function calls in `cuda::std::forward`

Also it makes reasoning about the code much more difficult compared to a simple operator that just takes a const reference

cub/device/dispatch/dispatch_reduce.cuh
miscco: Yeah, I just found the "no whitespace" option in the UI

cub/agent/agent_reduce.cuh
miscco: I think it would simplify the login here which is beneficial

cub/agent/agent_reduce.cuh
miscco: Urgh, yeah we are not comparing against the alignment of the pointer type but against 64bit alignment. Please disregard.

cub/thread/thread_operators.cuh
senior-zero: At some point we are going to alias these operators with libcu++ ones. If we do it later, we'd break behaviour. We can't do it right now, because libcu++ has to backport some specializations to previous dialects. So we'd have to wait till CUB 3.0 to make this change. It's not the case if we mimic the libcu++ behavior now. 

I'm also not sure what kind of throughput change you are referring to. There are no difference in generated [assembly](https://godbolt.org/z/T88xEhdnq) for cases when it doesn't affect behaviour.

miscco: Yeah sorry, I was referring to compiler throughput. instantiating functions is not entirely free.

cub/agent/agent_reduce.cuh
senior-zero: I'm not sure what you mean by login. We can keep the code as it was written and investigate the change you are referring to later.

senior-zero:
:warning: Breaking changes:

1. `AgentReduce`:
   1. now has extra template parameter representing the accumulator type.
   2. `ConsumeTile` uses accumulator type for thread aggregate instead of output iterator value type.
   3. `ConsumeTile` returns accumulator type instead of output iterator value type.
2. `DeviceReduceKernel` doesn't accept output iterator as a template parameter. Apart from that, it now accepts accumulator type.
3. `DeviceReduceSingleTileKernel` now accepts accumulator type.
4. `DeviceSegmentedReduceKernel` now accepts accumulator type.
5. `DeviceReducePolicy` now accepts accumulator type instead of input iterator value type. It also doesn't accept output iterator value type now.
6. `DispatchReduce`:
   1. now accepts accumulator type as last parameter. 
   2. now accepts initializer type instead of output iterator value type. 
   3. doesn't work with device extended lambdas now. 
   4. Constructor now accepts `init` as initial type instead of output iterator value type.
7. `DispatchSegmentedReduce`:
   1. accepts accumulator type as last parameter.
   2. accepts initializer type instead of output iterator value type. 
8. Thread operators now accept parameters using different types: `Equality`, `Inequality`, `InequalityWrapper`, `Sum`, `Difference`, `Division`, `Max`, `ArgMax`, `Min`, `ArgMin`.
9. `ThreadReduce` now accepts accumulator type and use different type for prefix
10. Default accumulator type is now selected by introspecting return type of the reduce operator.

allisonvacanti:
> Breaking changes

Thanks for writing this up! Can you edit the PR description/first comment and add this to it? I'm less likely to overlook it that way :)

senior-zero:
> > Breaking changes
> 
> Thanks for writing this up! Can you edit the PR description/first comment and add this to it? I'm less likely to overlook it that way :)

Sorry, didn't think about this aspect :smile: I'll definitely update the description.

PR #511, opened 2022-06-16
P2322R6 accumulator types for scan and reduce by key

senior-zero:
This PR addresses the following [issue](https://github.com/NVIDIA/cub/issues/428). It's based on the previous [PR](https://github.com/NVIDIA/cub/pull/509) so the first commit might be skipped during this review. 

:warning: Breaking changes:
1. Device scan, scan by key, and reduce by key now do not work with device extended lambdas by default. 
2. `AgentReduceByKey` now accepts accumulator type as the last template parameter.
3. `AgentScan` now accepts accumulator type as last template parameter.
4. `AgentScan::ConsumeRange` now accepts `prefix` as accumulator type instead of output iterator value type.
5. `AgentScanByKey` now accepts accumulator type as the last template parameter.
6. `DeviceReduceByKeyKernel` now accepts accumulator type as the last template parameter.
7. `DispatchReduceByKey` now accepts accumulator type as the last template parameter.
8. `DeviceScanKernel` now accepts accumulator type as the last template parameter.
9. `DeviceScanPolicy` now accepts accumulator type instead of output iterator value type.
10. `DispatchScan` now accepts accumulator type as a template parameter.
11. `DeviceScanByKeyKernel` now accepts accumulator type as a template parameter.
12. `DeviceScanByKeyPolicy` doesn't accept initial value type.
13. `DeviceScanByKeyPolicy` accepts accumulator type as a template parameter.
14. `DispatchScanByKey` now accepts accumulator type.

### DIFF
--- a/cub/agent/agent_reduce_by_key.cuh
+++ b/cub/agent/agent_reduce_by_key.cuh
@@ -134,7 +134,8 @@ template <typename AgentReduceByKeyPolicyT,
           typename NumRunsOutputIteratorT,
           typename EqualityOpT,
           typename ReductionOpT,
-          typename OffsetT>
+          typename OffsetT,
+          typename AccumT>
 struct AgentReduceByKey
 {
   //---------------------------------------------------------------------
@@ -151,19 +152,15 @@ struct AgentReduceByKey
   // The input values type
   using ValueInputT = cub::detail::value_t<ValuesInputIteratorT>;
 
-  // The output values type
-  using ValueOutputT =
-    cub::detail::non_void_value_t<AggregatesOutputIteratorT, ValueInputT>;
-
   // Tuple type for scanning (pairs accumulated segment-value with
   // segment-index)
-  using OffsetValuePairT = KeyValuePair<OffsetT, ValueOutputT>;
+  using OffsetValuePairT = KeyValuePair<OffsetT, AccumT>;
 
   // Tuple type for pairing keys and values
-  using KeyValuePairT = KeyValuePair<KeyOutputT, ValueOutputT>;
+  using KeyValuePairT = KeyValuePair<KeyOutputT, AccumT>;
 
   // Tile status descriptor interface type
-  using ScanTileStateT = ReduceByKeyScanTileState<ValueOutputT, OffsetT>;
+  using ScanTileStateT = ReduceByKeyScanTileState<AccumT, OffsetT>;
 
   // Guarded inequality functor
   template <typename _EqualityOpT>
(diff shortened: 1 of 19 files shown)


cub/detail/memory.cuh
```diff
@@ -0,0 +1,66 @@
...
+namespace detail
+{
+
+template <typename T,
+          typename U,
+          typename cuda::std::enable_if<
+            cuda::std::is_trivially_copyable<T>::value, 
+            int
+          >::type = 0>
+__host__ __device__ void uninitialized_copy(T *ptr, U &&val)
```
allisonvacanti, line 46: Consider renaming this header to `uninitialized_copy.cuh`. I tend to prefer more fine-grained headers (one feature per header) so builds can be more lean.

cub/detail/memory.cuh
```diff
@@ -0,0 +1,66 @@
...
+
+template <typename T, 
+         typename U,
+         typename cuda::std::enable_if<
+           !cuda::std::is_trivially_copyable<T>::value,
+           int
+         >::type = 0>
+__host__ __device__ void uninitialized_copy(T *ptr, U &&val)
+{
+  new (ptr) T(cuda::std::forward<U>(val));
```
allisonvacanti, line 59: There are several places in the previous commits that use placement-new directly. Should they be updated to use this?

allisonvacanti (approved):
Some comments inline. This will also need a breaking changes summary (APIs, behaviors) in the description.

cub/detail/memory.cuh
senior-zero: Could you specify these places?

senior-zero: I've found only the following places that existed before this PR:
```bash
:rg 'new\s*\('
cub/detail/uninitialized_copy.cuh
59:  new (ptr) T(cuda::std::forward<U>(val));

cub/block/block_load.cuh
364:        new(&items[ITEM]) InputT(block_itr[warp_offset + tid + (ITEM * CUB_PTX_WARP_THREADS)]);
401:            new(&items[ITEM]) InputT(block_itr[warp_offset + tid + (ITEM * CUB_PTX_WARP_THREADS)]);
```
I'm going to merge this. If there's anything I'm missing, I'll create a follow up PR.

allisonvacanti: I thought there were a few others, but may be wrong -- they might have just been placement-new + default ctor.

+1 to merging -- this definitely shouldn't hold anything up.

Issue #482, opened 2022-05-13
Single thread warp reduce


senior-zero:
Currently, we don't support single-threaded reductions in `cub::WarpReduce`. The documentation only states that `LOGICAL_WARP_THREADS` may be less than the number of hardware threads. The mentioned use case fits into this category. We should address this issue because the warp size might be a template parameter. Here's a reproducer that exposes compilation error:

```cpp
#include <cub/cub.cuh>

__global__ void kernel() {
  using warp_reduce_t = cub::WarpReduce<int, 1>;
  using storage_t = warp_reduce_t::TempStorage;

  __shared__ storage_t storage;
  warp_reduce_t(storage).Sum(1);
}

int main() {
  kernel<<<1, 1>>>();
}
```

```bash
:nvcc -gencode arch=compute_86,code=sm_86 main.cu
/usr/local/cuda/bin/../targets/x86_64-linux/include/cub/block/specializations/../../warp/specializations/warp_reduce_smem.cuh(69): error: shift count is negative
          detected during:
            instantiation of class "cub::WarpReduceSmem<T, LOGICAL_WARP_THREADS, PTX_ARCH> [with T=int, LOGICAL_WARP_THREADS=1, PTX_ARCH=0]" 
/usr/local/cuda/bin/../targets/x86_64-linux/include/cub/block/specializations/../../warp/warp_reduce.cuh(168): here
            instantiation of class "cub::WarpReduce<T, LOGICAL_WARP_THREADS, PTX_ARCH> [with T=int, LOGICAL_WARP_THREADS=1, PTX_ARCH=0]" 
main.cu(6): here

1 error detected in the compilation of "main.cu".
```

The compilation issue is caused by `1 << (STEPS - 1)` in cub/warp/specializations/warp_reduce_smem.cuh. Compilation is successful in the current version of cub because of the following change:
```cpp
    using InternalWarpReduce = cub::detail::conditional_t<                      
      IS_POW_OF_TWO, // => used to be (PTX_ARCH >= 300) && IS_POW_OF_TWO,                                       
      WarpReduceShfl<T, LOGICAL_WARP_THREADS, PTX_ARCH>,                        
      WarpReduceSmem<T, LOGICAL_WARP_THREADS, PTX_ARCH>>; 
```
[...]

Issue #279, opened 2021-03-31
Abstraction for Compute 8.0 warp reduce functions in CUB


pb-dseifert:
Hi Nvidia,
first, CUB has been a great library and has helped me achieve peak performance in my CUDA programs. Going forward, I'd like to use the warp reduction intrinsics introduced with Ampere. OTOH, I also need to keep my programs compatible with Pascal at a minimum, so getting the equivalents of `cub::WarpReduce::Sum()` for `__reduce_add_sync`, `__reduce_min_sync`, `__reduce_max_sync`, `__reduce_and_sync`, `__reduce_or_sync` and `__reduce_xor_sync`. While I could resort to `cub::WarpReduce::Reduce(T input, ReductionOp reduction_op)` for these, I doubt I would get the peak performance because `nvcc` is unlikely to optimize this whole bunch of code down into `__reduce_foo` when targetting Ampere. Is this a reasonable request?

allisonvacanti:
Thanks for the request.

This is something we are planning to look into. There is a lot of higher priority work to get through first, but exposing Ampere features through is high on our list.

PR #516, opened 2022-06-24
Fix main/GitHub/warp reduce

senior-zero:
This PR fixes the [issue](https://github.com/NVIDIA/cub/issues/482) of single-threaded warp reduces. It also addresses the [issue](https://github.com/NVIDIA/cub/issues/279) which requested hardware acceleration for reductions on ampere.

### DIFF
--- a/cub/warp/specializations/warp_reduce_shfl.cuh
+++ b/cub/warp/specializations/warp_reduce_shfl.cuh
@@ -492,6 +492,8 @@ struct WarpReduceShfl
         return output;
     }
 
+    // Warp reduce functions are not supported by nvc++ (NVBug 3694682)
+#ifndef _NVHPC_CUDA 
     template <class U = T>
     __device__ __forceinline__ 
     typename std::enable_if<
@@ -557,6 +559,7 @@ struct WarpReduceShfl
 
       return output;
     }
+#endif // _NVHPC_CUDA 
 
     /// Reduction
     template <
--- a/cub/warp/specializations/warp_reduce_shfl.cuh
+++ b/cub/warp/specializations/warp_reduce_shfl.cuh
@@ -39,6 +39,9 @@
 #include "../../util_type.cuh"
 
 #include <stdint.h>
+#include <type_traits>
+
+#include <nv/target>
 
 CUB_NAMESPACE_BEGIN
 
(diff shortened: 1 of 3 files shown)


cub/warp/specializations/warp_reduce_shfl.cuh
```diff
@@ -455,35 +458,121 @@ struct WarpReduceShfl
...
         T output = input;
 
-//        // Iterate reduction steps
-//        #pragma unroll
-//        for (int STEP = 0; STEP < STEPS; STEP++)
-//        {
-//            output = ReduceStep(output, reduction_op, last_lane, 1 << STEP, Int2Type<IsInteger<T>::IS_SMALL_UNSIGNED>());
-//        }
+        // Template-iterate reduction steps
+        ReduceStep(output, reduction_op, last_lane, Int2Type<0>());
```
brycelelbach, line 473: I find it a bit odd that `ReduceImpl` takes "all lanes valid" as as the first parameter, but it's the last parameter for `ReduceStep`. I'd prefer for them to be consistent.

cub/warp/specializations/warp_reduce_shfl.cuh
```diff
@@ -455,35 +458,121 @@ struct WarpReduceShfl
...
+
+        T output = input;
 
         // Template-iterate reduction steps
         ReduceStep(output, reduction_op, last_lane, Int2Type<0>());
 
         return output;
     }
 
+#ifndef _NVHPC_CUDA // NVBug 3694682
```
brycelelbach, line 495: Let's leave a more descriptive comment here - non-NVIDIA employees don't know what NVBUg 3694682. You can just say "NVC++ doesn't support the `__reduce_*_sync` primitives yet".

cub/warp/warp_reduce.cuh
```diff
@@ -603,6 +603,74 @@ public:
     //@}  end member group
 };
 
+template <typename T, int LEGACY_PTX_ARCH>
+class WarpReduce<T, 1, LEGACY_PTX_ARCH>
+{
+private:
+  using _TempStorage = cub::NullType;
+
+  _TempStorage &temp_storage;
```
brycelelbach, line 612: Let's try to remove this, we shouldn't need it, and it would be nice for this to be an empty class.

brycelelbach (approved):
Looks good to me.

Issue #522, opened 2022-07-02
Allow RDC builds without CDP support


senior-zero:
The following code leads to excessive memory footprint when compiled with `-rdc=true`:

```cpp
#include <iostream>
#include <cub/device/device_segmented_sort.cuh>

template class cub::DispatchSegmentedSort<false, int, int, int,int*, int*>;

int main() {
    size_t free_byte{}, total_byte{};
    double used{};

    cudaMemGetInfo(&free_byte, &total_byte);

    used = static_cast<double>(total_byte - free_byte) / 1024.0 / 1024.0;
    std::cout << "CUB " << CUB_VERSION << ": " << used << " MB" << std::endl;

    return 0;
}
```

```bash
:nvcc -O3 -rdc=true testSort.cu
:./a.out
CUB 200000: 1410.75 MB
:nvcc -O3 testSort.cu
:./a.out
CUB 200000: 616.75 MB
```

The issue can be reduced to the following code:
```cpp
#include <iostream>

__global__ void continuation() { }

__global__ void kernel()
{
  continuation<<<1, 256>>>();
  __cudaDeviceSynchronizeDeprecationAvoidance();
}

int main() {
  size_t free_byte{}, total_byte{};
  double used{};

  cudaMemGetInfo(&free_byte, &total_byte);

  used = static_cast<double>(total_byte - free_byte) / 1024.0 / 1024.0;
  std::cout << used << " MB" << std::endl;
}
```

The `__cudaDeviceSynchronizeDeprecationAvoidance` is used on device only when `debug_synchronous` is set. I suggest we restrict "synchronous" part if the option in the case of device-side usage. 

The second issue here is that we take `-rdc=true` as a desire to launch device-scope algorithms from device. Although `-rdc=true` is a requirement for device-side launches, it doesn't necessarily used for that purpose. The option might indicate that device code is distributed in a library, or that some device code is compiled in a different TU. I suggest we take a closer look at the overheads associated with device-side launches support and in case ones are found, separate `rdc` concept from device-side launch support.

allisonvacanti:
Agreed, this does sound like a reasonable option. We'd need to default to using the RDC state, but provide an opt-out for CDP support.

allisonvacanti:
Bumping priority after taking a closer look at this:

```
:nvcc -O3 -rdc=true testSort.cu
:./a.out
CUB 200000: 1410.75 MB
:nvcc -O3 testSort.cu
:./a.out
[...]

PR #523, opened 2022-07-02
Fix DebugSyncStream

senior-zero:
This PR addresses the following [issue](https://github.com/NVIDIA/cub/issues/522)

### DIFF
--- a/cub/util_device.cuh
+++ b/cub/util_device.cuh
@@ -538,8 +538,16 @@ CUB_RUNTIME_FUNCTION inline cudaError_t DebugSyncStream(cudaStream_t stream,
   }
 
 #if 1 // All valid targets currently support device-side synchronization
-  _CubLog("%s\n", "Synchronizing...");
-  return SyncStream(stream);
+  cudaError_t result = cudaSuccess;
+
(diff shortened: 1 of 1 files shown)


cub/util_device.cuh
```diff
@@ -538,8 +538,16 @@ CUB_RUNTIME_FUNCTION inline cudaError_t DebugSyncStream(cudaStream_t stream,
   }
 
 #if 1 // All valid targets currently support device-side synchronization
```
miscco, line 540: Should we also remove the conditional compilation and just use the new code unconditionally?

senior-zero: Good catch, thanks!

allisonvacanti: Leave this in -- the "dead code" here is a placeholder for an internal usecase.

allisonvacanti: Ignore that -- I missed that the new version still includes both paths 👍

allisonvacanti:
Marking as blocked pending #525.

senior-zero:
Addressed this differently in the following [PR](https://github.com/NVIDIA/cub/pull/527)

PR #524, opened 2022-07-07
Fix device-side debug synchronous usage for seg sort

senior-zero:
Backport of the following [fix](https://github.com/NVIDIA/cub/pull/523) to 1.17 branch.


leofang:
Q: Will we have another 1.17.x release?

PR #528, opened 2022-07-15
Disable managed memory for nvc++

miscco:
Using managed memory breaks certain tests like `test_thread_sort` at least on windows / WSL

Disable it uniformly to be on the safe side and let the user deside whether they want the fast footgun

### DIFF
--- a/cmake/CubBuildCompilerTargets.cmake
+++ b/cmake/CubBuildCompilerTargets.cmake
@@ -79,6 +79,7 @@ function(cub_build_compiler_targets)
     # * NVC++ accepts CUDA C++ in .cpp files but not .cu files.
     # TODO: This won't be necessary in the future.
     list(APPEND cxx_compile_options -cppsuffix=cu)
+    list(APPEND cxx_compile_options -gpu=nomanaged)
   endif()
 
   add_library(cub.compiler_interface INTERFACE)


senior-zero (approved):
This is an important change. It allows us treat nvcc and nvhpc builds with the same mindset. If someone develops in nvhpc environment and accidentally use pointer to the host memory, the'll notice issues only on CI stage. Also, is someone develop in linux environment, tests will behave nicely, but windows will segfault with difficult to find reasons (concurrent managed memory reads/writes with simultaneous kernel running).

allisonvacanti (approved):
This sounds reasonable to me. @dkolsen-pgi, do you see any potential issues from this change?

dkolsen-pgi:
The change looks fine to me.  I think the same effect could be achieved by changing `-stdpar` to `-cuda` in cmake/CubCompilerHacks.cmake.  The biggest difference between `-stdpar` and `-cuda` is that `-stdpar` turns on managed memory and `-cuda` does not.

This section right next to where the line was added is old.  The NVC++ part of the comment no longer applies.  This can be removed.
```
    # Today:
    # * NVCC accepts CUDA C++ in .cu files but not .cpp files.
    # * NVC++ accepts CUDA C++ in .cpp files but not .cu files.
    # TODO: This won't be necessary in the future.
    list(APPEND cxx_compile_options -cppsuffix=cu)
```

PR #535, opened 2022-07-21
Improve nvhpc configuration

### LINKED ISSUE #530: Actualize nvhpc configuration
Per @dkolsen-pgi [comment](https://github.com/NVIDIA/cub/pull/528#issuecomment-1185706011) we should stop treating `.cu` files as `.cpp` ones and use `-cuda` instead of `-stdpar + -gpu=nomanaged`.

miscco:
It was suggested that instead of treating .cu files as .cpp files and passing on `-stdpar` to nvc++ we should rather use the `-cuda` compile flag.

Also we found that nvc++ has different default settings with respect to denormals. So add them too.

Fixes #530

Note: I consolitated passing those flags into a single location. If we should keep them separate please hiss


cmake/CubCompilerHacks.cmake
```diff
@@ -25,7 +25,7 @@ if ("NVCXX" STREQUAL "${CMAKE_CUDA_COMPILER_ID}")
   endif()
 
   set(CMAKE_CXX_COMPILER "${CMAKE_CUDA_COMPILER}")
-  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -stdpar")
+  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -cuda -Mnodaz")
```
senior-zero, line 28: I now that managed memory is not used in this case, but I can see value in having `-gpu=nomanaged` written verbosely.  This way we don't rely on defaults and make our expectations visible. What do you think?

senior-zero (approved):

cmake/CubCompilerHacks.cmake
allisonvacanti: Was the managed memory issue WSL specific or did it affect native linux builds, too?

allisonvacanti: @miscco Can you move these back to the other file? I'd rather keep these properties on a target rather than polluting `CMAKE_CUDA_FLAGS`.

Also, I may be conflating issues with my WSL question above😅 Let me know if I'm not making sense there.

allisonvacanti:
Pinging @dkolsen-pgi to bless these nvc++ flag changes.

dkolsen-pgi:
I approve having the NVC++ options be `-cuda -Mnodaz`.  I have no opinion about which file should specify those options.

cmake/CubCompilerHacks.cmake
senior-zero: @allisonvacanti managed memory + concurrent GPU work + windows (or wsl) = segfault. It shouldn't affect linux builds.

miscco: Yeah this only affects windows. We might be able to better target that option but I am always a bit hesitant about diverging cmake configurations

miscco: @allisonvacanti I moved the options back and added a TODO for the managed memory option

allisonvacanti: Thanks! While we're updating this, could you move the `-cuda` option over to the other file, too? That should have been a target property all along.

We should also check whether Thrust needs these changes, too. The build systems are nearly identical, so it should be straightforward.

allisonvacanti (approved):
thanks! 👍

PR #538, opened 2022-07-25
Temporarily disable tests for non-RDC architectures.

allisonvacanti:
This works around some failures in Thrust's post-commit CI.

This is due to how `CMAKE_CUDA_FLAGS` works -- all targets in a directory
must share the same flags. Now that some tests always enable RDC to test
CDP launches, we have to disable the tests for non-RDC arches (53, 62, 72).

We can fix this properly once we migrate to using `CMAKE_CUDA_ARCHITECTURES`.

See comments in this patch for details.

### DIFF
--- a/cmake/CubCudaConfig.cmake
+++ b/cmake/CubCudaConfig.cmake
@@ -11,8 +11,11 @@ set(arch_message "CUB: Explicitly enabled compute architectures:")
 # reuse them if possible. After we transition to CMake 3.18 CUDA_ARCHITECTURE
 # target properties this will need to be updated.
 if (CUB_IN_THRUST)
-  # Configure to use all flags from thrust:
-  set(CMAKE_CUDA_FLAGS "${THRUST_CUDA_FLAGS_BASE} ${THRUST_CUDA_FLAGS_NO_RDC}")
+  # Configure to use all flags from thrust. See ThrustCudaConfig.cmake for
+  # details.
+  set(CUB_CUDA_FLAGS_BASE "${THRUST_CUDA_FLAGS_BASE}")
+  set(CUB_CUDA_FLAGS_RDC "${THRUST_CUDA_FLAGS_RDC}")
+  set(CUB_CUDA_FLAGS_NO_RDC "${THRUST_CUDA_FLAGS_NO_RDC}")
 
   # Update the enabled architectures list from thrust
   foreach (arch IN LISTS all_archs)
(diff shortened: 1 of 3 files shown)


allisonvacanti:
This fixes failures in Thrust's post-commit CI, where we were building tests that depend on RDC for arches that don't support it.

This is a temporary work-around -- we'll have to wait until we're using `CMAKE_CUDA_ARCHITECTURES` to fix this properly.

allisonvacanti:
gpuCI: NVIDIA/thrust#1755

senior-zero (changes requested):
We seem to need to compile targets that doesn't require RDC for 53, 62 and 72 architectures. Right now we just skip these architectures completely for our tests.

allisonvacanti:
> Right now we just skip these architectures completely for our tests.

We'll have to skip those tests (temporarily) until we can start using the CMake `CUDA_ARCHITECTURES` property. Hopefully we'll remove the remaining blockers for that work soon.

That said, I do need to make sure to update the flags for the example builds. Update incoming!

cmake/CubCudaConfig.cmake
```diff
@@ -11,8 +11,11 @@ set(arch_message "CUB: Explicitly enabled compute architectures:")
 # reuse them if possible. After we transition to CMake 3.18 CUDA_ARCHITECTURE
 # target properties this will need to be updated.
 if (CUB_IN_THRUST)
-  # Configure to use all flags from thrust:
-  set(CMAKE_CUDA_FLAGS "${THRUST_CUDA_FLAGS_BASE} ${THRUST_CUDA_FLAGS_NO_RDC}")
+  # Configure to use all flags from thrust. See ThrustCudaConfig.cmake for
+  # details.
+  set(CUB_CUDA_FLAGS_BASE "${THRUST_CUDA_FLAGS_BASE}")
```
miscco, line 16: Nitpick: I do not like `BASE`, would it be possible to change that to `CUB_CUDA_FLAGS_COMMON`

miscco (approved):

cmake/CubCudaConfig.cmake
allisonvacanti: Good news :) We'll eventually rip all of these out when we are able to migrate to using the `CUDA_ARCHITECTURE` target property. I'll leave these as-is since these variables are just temporary.

senior-zero (approved):

PR #546, opened 2022-08-09
Fix libcu++ namespace

senior-zero:
This PR harmonizes libcu++ namespace usage with the following [PR](https://github.com/NVIDIA/thrust/pull/1763).


miscco (approved):
I am slightly inclined to define a macro for that to ensure that we a) never forget it again and b) do not have to type so much.

That said I am happy with this PR so no change requested

allisonvacanti (approved):
Doesn't need to block this fix, but we should add a check for this to the `CheckSourceFiles.cmake` test.

PR #549, opened 2022-08-13
Fix BlockReduceRaking non-commutative docs

upsj:
The documentation contradicts itself here: BlockReduceRaking seems to support non-commutative operations, but BlockReduceRakingCommutativeOnly doesn't


senior-zero (approved):

senior-zero:
Thank you for noticing this!

PR #553, opened 2022-08-18
Deprecate CTK<9

senior-zero:
:warning: Deprecated `CUB_USE_COOPERATIVE_GROUPS`

Currently, CUB has code specializations for CTK < 9. Based of the CTK version, we provide a `CUB_USE_COOPERATIVE_GROUPS` macro. If the macro is absent, we disables `*_sync` variants of PTX instructions / intrinsics. This PR removes the macro and all associated code specializations. We don't test that and I have doubts if the following even legal for all the CUB facilities:
```cpp
__device__  __forceinline__ void WARP_SYNC(unsigned int member_mask)
{
#ifdef CUB_USE_COOPERATIVE_GROUPS
    __syncwarp(member_mask);
#endif
}
```

### DIFF
--- a/cub/util_arch.cuh
+++ b/cub/util_arch.cuh
@@ -44,12 +44,6 @@ CUB_NAMESPACE_BEGIN
 
 #ifndef DOXYGEN_SHOULD_SKIP_THIS    // Do not document
 
-#if ((__CUDACC_VER_MAJOR__ >= 9) || defined(_NVHPC_CUDA) ||            \
-     CUDA_VERSION >= 9000) &&                                                  \
-  !defined(CUB_USE_COOPERATIVE_GROUPS)
-#define CUB_USE_COOPERATIVE_GROUPS
-#endif
-
 /// In device code, CUB_PTX_ARCH expands to the PTX version for which we are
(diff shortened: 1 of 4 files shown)


cub/util_arch.cuh
```diff
@@ -44,12 +44,6 @@ CUB_NAMESPACE_BEGIN
 
 #ifndef DOXYGEN_SHOULD_SKIP_THIS    // Do not document
 
-#if ((__CUDACC_VER_MAJOR__ >= 9) || defined(_NVHPC_CUDA) ||            \
-     CUDA_VERSION >= 9000) &&                                                  \
-  !defined(CUB_USE_COOPERATIVE_GROUPS)
-#define CUB_USE_COOPERATIVE_GROUPS
```
allisonvacanti, line 50: Leave this defined unconditionally for now and add a note stating that it's deprecated, since users may depend on it. 

Mark this as a breaking change and put a note in the PR description stating that this macro is deprecated. We can fully remove it after a few releases 👍

allisonvacanti (approved):
LGTM, with a minor change pointed out inline.

PR #558, opened 2022-08-22
Add clang-format

senior-zero:
This PR adds initial version of style guide in the form of clang-format file. The guide is [copyed](https://github.com/NVIDIA/nvbench/blob/main/.clang-format) from the nvbench repo.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,76 @@
+BasedOnStyle: LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true
+AlignEscapedNewlines: Right
+AlignOperands: true
+AllowAllArgumentsOnNextLine: false
+AllowAllConstructorInitializersOnNextLine: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: false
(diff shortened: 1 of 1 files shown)


jrhemstad (approved):

.clang-format
```diff
@@ -0,0 +1,76 @@
...
+PenaltyBreakBeforeFirstCallParameter: 50
+PenaltyBreakComment: 0
+PenaltyBreakFirstLessLess: 0
+PenaltyBreakString: 70
+PenaltyBreakTemplateDeclaration: 0
+PenaltyExcessCharacter: 100
+PenaltyReturnTypeOnItsOwnLine: 90
+PointerAlignment: Right
+ReflowComments: true
+SortIncludes: CaseInsensitive
```
miscco, line 58: Note to future self. There is `IncludeBlocks` and `IncludeCategories` that can be used to automagically create sorted blocks of different types of includes, that are matched via a regex. This might be useful later on for example in `thrust` to differentiate the different dependencies

senior-zero: Wow, that would be awesome to have!

miscco: Shameless plug of the MSVC STL https://github.com/microsoft/STL/blob/main/.clang-format#L117-L131

PR #562, opened 2022-08-27
Fix wrong tid when blockDim.y or z > 1

sjfeng1999:


senior-zero:
@sjfeng1999 thank you! I've updated the tests to capture the behaviour you fixed. I'll run the CI now.

senior-zero (approved):

PR #565, opened 2022-08-31
Add DeviceMergeSort::StableSortKeysCopy API

davidwendt:
CUB has a  `SortKeysCopy` and a `StableSortKeys`. This PR adds `StableSortKeysCopy` so a sort in-place is not required. The in-place `StableSortKeys`  requires a temporary vector if a copy is needed.

Since `StableSortKeys` calls `SortKeys` and `SortKeysCopy` uses the same kernels as `SortKeys` (which appear to already be stable-sort enabled), the new `StableSortKeysCopy` simply calls `SortKeysCopy`.

### DIFF
--- a/cub/device/device_merge_sort.cuh
+++ b/cub/device/device_merge_sort.cuh
@@ -34,7 +34,6 @@
 
 CUB_NAMESPACE_BEGIN
 
-
 /**
  * @brief DeviceMergeSort provides device-wide, parallel operations for
  *        computing a merge sort across a sequence of data items residing within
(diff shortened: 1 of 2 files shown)


cub/device/device_merge_sort.cuh
```diff
@@ -994,6 +994,152 @@ struct DeviceMergeSort
...
+
+    return StableSortKeysCopy<KeyInputIteratorT, KeyIteratorT, OffsetT, CompareOpT>(d_temp_storage,
+                                                                                    temp_storage_bytes,
+                                                                                    d_input_keys,
+                                                                                    d_output_keys,
+                                                                                    num_items,
+                                                                                    compare_op,
+                                                                                    stream);
+  }
+
```
senior-zero, line 1142:
```suggestion
```

This is not required. The `debug_synchronous` overloads exist so as not to break API. For the new functions, we don't have anything to break, so let's remove this overload.

test/test_device_merge_sort.cu
```diff
@@ -234,6 +234,18 @@ void TestKeys(std::int64_t num_items,
     CustomLess()));
 
   AssertTrue(CheckResult(d_keys));
+
+  CubDebugExit(cub::DeviceMergeSort::StableSortKeysCopy(
```
senior-zero, line 238: Just realized that there might be a value in something like:
```suggestion
  thrust::fill(d_keys.begin(), d_keys.end(), KeyType{});
  CubDebugExit(cub::DeviceMergeSort::StableSortKeysCopy(
```

senior-zero (approved):
Thank you for the contribution! A few minor suggestions below.

miscco (approved):
Thanks for the contribution and the excellent documentation

PR #571, opened 2022-09-08
Don't use char in tests

senior-zero:
There are compilation issues for our tests on some platforms due to `char` usage. According to the standard:
> Type char is a distinct type that has an implementation-defined choice of “signed char” or “unsigned char” as its underlying type.

CUDA vector types add to this confusion by using `char2` as a synonym for `signed char` instead of `char`. Ideally, we'd use `std::int8_t` and `std::uint8_t` here, but I'd like to postpone this refactoring, since the process of replacing primitive types with `std::*` ones in already in progress ([Catch2 PR](https://github.com/NVIDIA/cub/pull/365)). As a quick fix, I use signed char when applicable for now.

### DIFF
--- a/test/test_block_load_store.cu
+++ b/test/test_block_load_store.cu
@@ -517,7 +517,7 @@ int main(int argc, char** argv)
 
     // Compile/run thorough tests
 #if TEST_VALUE_TYPES == 0
-    TestThreads<char>(2, 0.8f);
+    TestThreads<signed char>(2, 0.8f);
     TestThreads<int>(2, 0.8f);
     TestThreads<long>(2, 0.8f);
 #elif TEST_VALUE_TYPES == 1
--- a/test/test_util.h
+++ b/test/test_util.h
@@ -1134,7 +1134,7 @@ std::ostream& operator<<(std::ostream& os, const CUB_NS_QUALIFIER::KeyValuePair<
 /**
  * Define for types
  */
-CUB_VEC_OVERLOAD(char, char)
+CUB_VEC_OVERLOAD(char, signed char)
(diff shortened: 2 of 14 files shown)


test/test_block_load_store.cu
```diff
@@ -517,7 +517,7 @@ int main(int argc, char** argv)
 
     // Compile/run thorough tests
 #if TEST_VALUE_TYPES == 0
-    TestThreads<char>(2, 0.8f);
```
miscco, line 520: I am slightly surprised that we only test signed integers but that is definitely preexisting, so fine by me

test/test_util.h
```diff
@@ -1134,7 +1134,7 @@ std::ostream& operator<<(std::ostream& os, const CUB_NS_QUALIFIER::KeyValuePair<
 /**
  * Define for types
  */
-CUB_VEC_OVERLOAD(char, char)
+CUB_VEC_OVERLOAD(char, signed char)
```
miscco, line 1137: Are you missing the first `char`?
```suggestion
CUB_VEC_OVERLOAD(signed char, signed char)
```

senior-zero: No, it's a name that is going to be turned into `char2`, `char4` by the macro.

test/test_block_load_store.cu
senior-zero: For load facility, I don't think it's reasonable to test multiple types of the same size.

test/test_util.h
miscco: urgh, totally not tricky

miscco (approved):

jrhemstad (approved):

PR #582, opened 2022-10-17
Fix reordering in CUB member initializer lists

senior-zero:
This PR fixes reordering in member initializer lists. Since `-Wall` was added to host compiler, kernels were not covered. Specifying `-Wreorder` directly led to having those reordering visible.

### DIFF
--- a/cmake/CubBuildCompilerTargets.cmake
+++ b/cmake/CubBuildCompilerTargets.cmake
@@ -9,6 +9,7 @@
 function(cub_build_compiler_targets)
   set(cxx_compile_definitions)
   set(cxx_compile_options)
+  set(cuda_compile_options)
 
   if ("MSVC" STREQUAL "${CMAKE_CXX_COMPILER_ID}")
     append_option_if_available("/W4" cxx_compile_options)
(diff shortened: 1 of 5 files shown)


cmake/CubBuildCompilerTargets.cmake
```diff
@@ -38,6 +39,7 @@ function(cub_build_compiler_targets)
     # Some tests require /bigobj to fit everything into their object files:
     append_option_if_available("/bigobj" cxx_compile_options)
   else()
+    append_option_if_available("-Wreorder" cuda_compile_options)
```
miscco, line 42: Could we please separate that one from the others.

It is really easy to overlook that one is `cuda_compile_options` and the others `cxx_compile_options`

miscco (approved):

cmake/CubBuildCompilerTargets.cmake
```diff
@@ -90,7 +92,7 @@ function(cub_build_compiler_targets)
       # CMake can't split genexs, so this can't be formatted better :(
       # This is:
       # if (using CUDA and CUDA_COMPILER is NVCC) add -Xcompiler=opt:
-      $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:-Xcompiler=${cxx_option}>
+      $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:${cuda_compile_options} -Xcompiler=${cxx_option}>
```
allisonvacanti, line 95: This will specify `-Wreorder` multiple times, once for each `cxx_option`. This should be done outside of the `foreach` loop.

PR #588, opened 2022-11-01
Fix warp scan test

senior-zero:
Warp scan is not expected to be used with non-power-of-two logic warp sizes when there's more than one logic warp per hardware warp. Our test happened to be relying on that. This PR fixes warp scan test not to launch more threads than warp reduce expects.

### DIFF
--- a/test/test_warp_scan.cu
+++ b/test/test_warp_scan.cu
@@ -1,6 +1,6 @@
 /******************************************************************************
  * Copyright (c) 2011, Duane Merrill.  All rights reserved.
- * Copyright (c) 2011-2018, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 2011-2022, NVIDIA CORPORATION.  All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
@@ -47,7 +47,7 @@ using namespace cub;
 // Globals, constants and typedefs
 //---------------------------------------------------------------------
 
-static const int        NUM_WARPS       = 2;
+static const int        MAX_WARPS       = 2;
 
 
 bool                    g_verbose       = false;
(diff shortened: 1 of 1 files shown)


test/test_warp_scan.cu
```diff
@@ -399,7 +401,10 @@ void Test(
     InitialValueT   initial_value)
 {
     enum {
-        TOTAL_ITEMS = LOGICAL_WARP_THREADS * NUM_WARPS,
+        IS_ARCH_WARP = (LOGICAL_WARP_THREADS == CUB_WARP_THREADS(0)),
```
kaatish, line 404: What is CUB_WARP_THREADS supposed to do?

test/test_warp_scan.cu
```diff
@@ -458,7 +464,7 @@ void Test(
     fflush(stdout);
 
     // Run aggregate/prefix kernel
-    WarpScanKernel<LOGICAL_WARP_THREADS, TEST_MODE><<<1, TOTAL_ITEMS>>>(
+    WarpScanKernel<LOGICAL_WARP_THREADS, TEST_MODE><<<1, LOGICAL_WARP_THREADS * TOTAL_WARPS>>>(
```
kaatish, line 467: Why was this change made? Isn't `TOTAL_ITEMS == LOGICAL_WARP_THREADS * TOTAL_WARPS`

test/test_warp_scan.cu
senior-zero: It represents hardware warp size throughout the project.

test/test_warp_scan.cu
senior-zero: It is, but items being processed doesn't immediately represent the number of threads you want. Now the code is written in terms of threads, not items.

kaatish (approved):

Issue #584, opened 2022-10-25
Support reduce and scan for more than 2^31 items


milthorpe:
`cub::DeviceReduce` and `cub::DeviceScan` specify the input size as `int num_items`, which limits reductions/scans to 2^31-1 items. Other CUB functions have the type of the number of items as a template parameter e.g. `cub::DeviceRadixSort` accepts `NumItemsT num_items`. 

The `num_items` parameter should be replaced with a templated parameter `NumItemsT num_items` for both `cub::DeviceReduce` and `cub::DeviceScan`, to allow reductions and scans of more than 2^31 items.

A simple test code for `DeviceReduce` with `size_t num_items` would be as follows:

```
#include <cub/cub.cuh>

using namespace cub;

template <typename T> void cubSum(const T *d_in, T *d_out, size_t num_items) {
  void *d_temp_storage = NULL;
  size_t temp_storage_bytes = 0;
  CubDebugExit(DeviceReduce::Sum(d_temp_storage, temp_storage_bytes, d_in, d_out, num_items));
  cudaMalloc(&d_temp_storage, temp_storage_bytes);

  CubDebugExit(DeviceReduce::Sum(d_temp_storage, temp_storage_bytes, d_in, d_out, num_items));

  cudaFree(d_temp_storage);
}

int main(int argc, char* argv[]) {
    size_t num_items = 2l<<30;
    double *h_in = new double[num_items];
    for (size_t i=0; i<num_items; i++) h_in[i] = i;
    double result = 0;

    CachingDeviceAllocator  g_allocator(true);

    double *d_in = NULL;
    CubDebugExit(g_allocator.DeviceAllocate((void**)&d_in, sizeof(double) * num_items));

    // Initialize device input
    CubDebugExit(cudaMemcpy(d_in, h_in, sizeof(double) * num_items, cudaMemcpyHostToDevice));

    double *d_out = NULL;
    CubDebugExit(g_allocator.DeviceAllocate((void**)&d_out, sizeof(double)));

    cubSum(d_in, d_out, num_items);

    CubDebugExit(cudaMemcpy(&result, d_out, sizeof(double), cudaMemcpyDeviceToHost));
    printf("result: %e\n", result);

    if (h_in) delete[] h_in;
    if (d_in) CubDebugExit(g_allocator.DeviceFree(d_in));
    if (d_out) CubDebugExit(g_allocator.DeviceFree(d_out));
}
```

milthorpe:
[...]

PR #589, opened 2022-11-01
Support reduction for more than 2^31 items

senior-zero:
This PR partially addresses the following [issue](https://github.com/NVIDIA/cub/issues/584). Supporting larger offsets for `ArgMin` and `ArgMax` would have to wait, since it's a bit more involving. There also could be a bit better testing, but it might wait till Catch2 rewrite. 

The main purpose of this PR is to unblock original issue by providing an example of what we expect from such a change. The main direction of this modification can be found [here](https://github.com/NVIDIA/cub/issues/212).

### DIFF
--- a/cub/device/device_reduce.cuh
+++ b/cub/device/device_reduce.cuh
@@ -38,6 +38,7 @@
 #include <limits>
 
 #include <cub/config.cuh>
+#include <cub/detail/choose_offset.cuh>
 #include <cub/device/dispatch/dispatch_reduce.cuh>
 #include <cub/device/dispatch/dispatch_reduce_by_key.cuh>
 #include <cub/iterator/arg_index_input_iterator.cuh>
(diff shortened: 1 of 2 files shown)


elstehle (approved):
LGTM

kaatish (approved):

senior-zero:
Because `ChooseOffsetT` leads to unsigned offsets usage the codegen changed (14% increase in kernel size). Benchmarks show about 1% slowdown for reduction due to that. I've changed `ChooseOffsetT` to provide singed offsets instead, which leads to the same code generated. Also benchmarked radix sort, since it's affected by the change and got up to 4% speedup there.

leofang:
Hi @senior-zero, glad to see this patch landed! Would it be backported to the 1.17.x branch?

senior-zero:
Hello @leofang! I completely forgot to reply about backporting this PR. Sorry about that!

I wouldn't like to backport this. Is there something stopping you from using the newer version of CUB. Please let us know if we can help switch to a more recent version of CUB or fix potential issues.

PR #590, opened 2022-11-08
Correct install(DIRECTORY ... PATTERN) which matches at end of file name

robertmaynard:


cmake/CubInstallRules.cmake
```diff
@@ -18,7 +18,7 @@ install(DIRECTORY "${CUB_SOURCE_DIR}/cub"
 
 install(DIRECTORY "${CUB_SOURCE_DIR}/cub/cmake/"
   DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/cub"
-  PATTERN cub-header-search EXCLUDE
```
senior-zero, line 21: @robertmaynard Could you please explain why this is needed in the installation at all? As I understand, this pattern installs `cub-header-search.cmake.in` which is used below anyways to generate `cub-header-search.cmake` on the installation side. I guess we can just remove this.

senior-zero (changes requested):

cmake/CubInstallRules.cmake
robertmaynard: The PATTERN <tail-glob> EXCLUDE removes items from the set to install. 

In this case we are saying install everything from this directory except files end in .cmake.in. The problem this PR solves is that PATTERN looks at the end of file(+extension) so our previous logic was wrong.

senior-zero: Got it, thanks for the clarification!

senior-zero (approved):

PR #592, opened 2022-11-15
Fix overflow in reduce

senior-zero:


cub/agent/agent_reduce.cuh
```diff
@@ -355,7 +355,7 @@ struct AgentReduce
   {
     AccumT thread_aggregate{};
 
-    if (even_share.block_offset + TILE_ITEMS > even_share.block_end)
+    if (even_share.block_end - even_share.block_offset < TILE_ITEMS)
```
canonizer, line 358: To clarify my assumptions: is it always true that `even_share.block_end <= even_share.block_offset`? Can they be equal?

cub/agent/agent_reduce.cuh
```diff
@@ -377,14 +377,17 @@ struct AgentReduce
     even_share.block_offset += even_share.block_stride;
 
     // Consume subsequent full tiles of input
-    while (even_share.block_offset + TILE_ITEMS <= even_share.block_end)
+    while (even_share.block_offset <= even_share.block_end - TILE_ITEMS)
```
canonizer, line 380: What happens if `even_share.block_end < TILE_ITEMS`? Won't the following check be better:

```
while (even_share.block_end - even_share.block_offset >= TILE_ITEMS)
```

cub/agent/agent_reduce.cuh
```diff
@@ -377,14 +377,17 @@ struct AgentReduce
...
     {
       ConsumeTile<false>(thread_aggregate,
                          even_share.block_offset,
                          TILE_ITEMS,
                          Int2Type<true>(),
                          can_vectorize);
-      even_share.block_offset += even_share.block_stride;
+      const OffsetT new_offset = even_share.block_offset + even_share.block_stride;
+      even_share.block_offset = (new_offset > even_share.block_offset)
```
canonizer, line 388: Won't the following be better:

```
if (even_share.block_end - even_share.block_offset > even_share.block_stride)
{
    even_share.block_offset += even_share.block_stride;
}
```

cub/agent/agent_reduce.cuh
senior-zero: In my understanding, the only condition in which we get `even_share.block_end < TILE_ITEMS` is when there's one CTA per range. In this case, we should never reach this code and instead exists the function in the check above:
```cpp
if (even_share.block_end - even_share.block_offset < TILE_ITEMS)
{
  // First tile isn't full (not all threads have valid items)
  int valid_items = even_share.block_end - even_share.block_offset;
  ConsumeTile<true>(thread_aggregate,
                    even_share.block_offset,
                    valid_items,
                    Int2Type<false>(),
                    can_vectorize);
  return BlockReduceT(temp_storage.reduce)
    .Reduce(thread_aggregate, reduction_op, valid_items);
}
```
The motivation behind `while (even_share.block_offset <= even_share.block_end - TILE_ITEMS)` was that this way we can sum offsets (`even_share.block_end - TILE_ITEMS`) once per invocation, instead once per iteration in the `while (even_share.block_end - even_share.block_offset >= TILE_ITEMS)` case. 

Is there any other ways in which the proposed code is better?

elstehle: I think we would not run into the underflow case, where `even_share.block_end < TILE_ITEMS`, because we would return early by branching into this conditional (above):
```
if (even_share.block_end - even_share.block_offset < TILE_ITEMS)
```

Nonetheless, I think @canonizer 's suggestion would be preferable here.

cub/agent/agent_reduce.cuh
miscco: I believe It would be best to invert the condition
```cpp
if (even_share.block_end - even_share.block_offset < even_share.block_stride)
{
    early return here
}
even_share.block_offset += even_share.block_stride;
...
```

cub/agent/agent_reduce.cuh
miscco: I am not happy, that we transform the condition differently here and below. I like @canonizer suggestion below

test/test_device_reduce.cu
```diff
@@ -1360,10 +1360,10 @@ void TestBigIndicesHelper(int magnitude)
 template <typename T>
 void TestBigIndices()
 {
-  TestBigIndicesHelper<T>(30);
-  TestBigIndicesHelper<T>(31);
-  TestBigIndicesHelper<T>(32);
-  TestBigIndicesHelper<T>(33);
+  TestBigIndicesHelper<T, std::uint32_t>(1ull << 30);
```
miscco, line 1363: What is the confidence we gain from 30 as opposed to 31?

Are there different code paths?

cub/agent/agent_reduce.cuh
senior-zero: @miscco, could you elaborate on why it is better?

cub/agent/agent_reduce.cuh
senior-zero: @miscco, @canonizer suggestion doesn't change the fact that this line, or the line below has to be changed.

cub/agent/agent_reduce.cuh
miscco:
```cpp
const OffsetT new_offset = even_share.block_offset + even_share.block_stride;
```
Could potentially overflow afaik. Also I would believe, that it is generally always better to early return

senior-zero: @miscco the potential overflow is handled below. Early exit adds about 7% to the generated kernel size, so I was suspicious about the suggestion. I've benchmarked the change and there seems to be no huge difference in performance. Since I don't have preferences here, will trust your judgement.

cub/agent/agent_reduce.cuh
senior-zero: @canonizer, for segmented reduce `block_offset` is always less than `block_end`. For reduce the number of blocks is about `RoundUp(num_items, tile_size)`, while `block_offset` is just `block_id * TILE_ITEMS` and `block_end` is `num_items`. The case of `num_items == 0` is treated differently, so I don't think `block_end` can be equal to `block_offset`. Could you elaborate on why it's relevant here?

cub/agent/agent_reduce.cuh
miscco: Urgh, then please leav this as is. If there is no difference but a regression in kernel size I am all fine with keeping it as is

cub/agent/agent_reduce.cuh
senior-zero: Benchmarked the difference between `while (even_share.block_offset <= even_share.block_end - TILE_ITEMS)` and `while (even_share.block_end - even_share.block_offset >= TILE_ITEMS)`. It seems that my motivation above is verified. Got up to 5% stable slowdown for `while (even_share.block_end - even_share.block_offset >= TILE_ITEMS)` on int16_t and uint32_t. Will go with `while (even_share.block_offset <= even_share.block_end - TILE_ITEMS)`.

cub/agent/agent_reduce.cuh
miscco: I am wondering whether the whole algorithm would be simpler if we used `int valid_items = even_share.block_end - even_share.block_offset;` as the main variable instead of repeatedly computing the remaining number of items.

That said, the change is definitely correct and a large scale refactor is a bit too much right now

cub/agent/agent_reduce.cuh
elstehle: Interesting! Good to know. Thanks for checking.

canonizer: It looks like at this point, it holds that `even_share.block_end >= TILE_ITEMS`, so there won't be any underflow. If you want to keep the code as is, it's a good idea to add a comment.

I'd personally prefer `even_share.block_end - even_share.block_offset >= TILE_ITEMS` for uniformity, but it's up to you.

canonizer (approved):
Approved, provided that the comments are addressed.

cub/agent/agent_reduce.cuh
senior-zero: > I'd personally prefer even_share.block_end - even_share.block_offset >= TILE_ITEMS for uniformity, but it's up to you.

I believe it's a matter of performance rather than preference. [Here](https://github.com/NVIDIA/cub/pull/592#discussion_r1027374025) I share 5% regression from this suggestion. And at the beginning of this thread is the intuition for the source of this regression.

miscco (approved):

Issue #594, opened 2022-11-21
Invalid result from DeviceSegmentedSort::SortPairs/SortKeys when keys are bool type


davidwendt:
If the keys parameter is type `bool` then `DeviceSegmentedSort::SortPairs` and `DeviceSegmentedSort::SortKeys` return invalid results. The following example reproduces the error consistently. The error occurs with `SortKeys` as well but is easier to see with `SortPairs`

```
#include <thrust/device_vector.h>
#include <cub/cub.cuh>
#include <cuda_runtime.h>
#include <vector>
#include <iostream>

template <typename T>
std::vector<int> segmented_sort(bool const *h_data, size_t input_size,
                                int const *h_segs, size_t segs_size)
{
  T *d_data;
  cudaMalloc(&d_data, input_size * sizeof(bool));
  cudaMemcpy(d_data, h_data, input_size * sizeof(bool), cudaMemcpyHostToDevice);

  thrust::device_vector<int> d_segs(segs_size);
  cudaMemcpy(d_segs.data().get(), h_segs, segs_size * sizeof(int), cudaMemcpyHostToDevice);

  T *d_output;
  cudaMalloc(&d_output, input_size * sizeof(bool));

  thrust::device_vector<int> d_indices(input_size);
  thrust::sequence(thrust::device, d_indices.begin(), d_indices.end(), 0);

  void *d_temp_storage = nullptr;
  size_t temp_storage_bytes = 0;
  cub::DeviceSegmentedSort::SortPairs(d_temp_storage, temp_storage_bytes,
                                      d_data, d_output,
                                      d_indices.data().get(), d_indices.data().get(),
                                      input_size, d_segs.size() - 1,
                                      d_segs.begin(), d_segs.begin() + 1);

  cudaMalloc(&d_temp_storage, temp_storage_bytes);

  cub::DeviceSegmentedSort::SortPairs(d_temp_storage, temp_storage_bytes,
                                      d_data, d_output,
                                      d_indices.data().get(), d_indices.data().get(),
                                      input_size, d_segs.size() - 1,
                                      d_segs.begin(), d_segs.begin() + 1);

  cudaFree(d_temp_storage);
  cudaFree(d_data);
  cudaFree(d_output);

  std::vector<int> h_indices(d_indices.size());
  cudaMemcpy(h_indices.data(), d_indices.data().get(), d_indices.size() * sizeof(int), cudaMemcpyDeviceToHost);
  return h_indices;
}

int main()
{
[...]

PR #595, opened 2022-11-22
Fix seg sort for bool keys

senior-zero:
This PR addresses the following [issue](https://github.com/NVIDIA/cub/issues/594) and enables `bool` as key type in the new device segmented sort. The issue happens only for small segments since that's what is triggering merge sort to be used. Merge sort relies on comparison and we end up comparing `true` or `false` with `oob = 0xFF`. Initially, I was trying to `Traits<KeyT>::TwiddleIn(key)` once before sorting and `Traits<KeyT>::TwiddleOut(key)` after sorting so as to get consistency with what radix counterpart is doing. Although this simplifies the code (no need to treat `__half` etc.), this also introduces noticeable slowdown (4%) into sorting of small segments on signed key types. The slowdown disappears only on `std::int64_t`. So, instead we change the way `oob` is computed. This change doesn't affect performance. 

There was also a suggestion to change `Traits<bool>::MAX_KEY` to be `true` instead of `0xFF`, but I'm afraid this is a silently breaking change. Currently, expectations of `Traits<bool>::MAX_KEY` is to get `0b1..1` which will change if we go this route.

### DIFF
--- a/cub/agent/agent_sub_warp_merge_sort.cuh
+++ b/cub/agent/agent_sub_warp_merge_sort.cuh
@@ -155,6 +155,25 @@ class AgentSubWarpSort
     return lhs == rhs;
   }
 
+  __device__ static bool get_oob_default(Int2Type<true> /* is bool */) 
+  {
+    // Traits<KeyT>::MAX_KEY for `bool` is 0xFF which is different from `true` and makes
+    // comparison with oob unreliable.
+    return !IS_DESCENDING;
+  }
+
+  __device__ static KeyT get_oob_default(Int2Type<false> /* is bool */) 
+  {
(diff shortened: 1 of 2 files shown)


elstehle (approved):
Looks good 👍 

> There was also a suggestion to change Traits<bool>::MAX_KEY to be true instead of 0xFF, but I'm afraid this is a silently breaking change. Currently, expectations of Traits<bool>::MAX_KEY is to get 0b1..1 which will change if we go this route.

Agree, changing `Traits<bool>::MAX_KEY` would change the behaviour of the radix sort, if there were `bool` items different from `true` and `false`.

miscco (approved):

PR #597, opened 2022-11-25
Fix device-side reduce launch with large inputs

senior-zero:
When there are more than one tile of input items, we launch two kernels for reduction. During the second kernel launch, we accumulate partial aggregates from thread blocks. In this case, `int reduce_grid_size` is passed to `OffsetT num_items` kernel parameter. Thrust `triple_chevron` is not type safe when used on device. These facts lead to a mismatch of data size we put into a param buffer (4 bytes) and that the kernel reads from the buffer (8 bytes). This seems to be a bug that was long present in thrust, that uses dispatch layer directly with offset types different from `int32_t`. In the case of host-side usage, type conversion takes place, leading to normal execution.

### DIFF
--- a/cub/device/dispatch/dispatch_reduce.cuh
+++ b/cub/device/dispatch/dispatch_reduce.cuh
@@ -783,7 +783,7 @@ struct DispatchReduce : SelectedPolicy
         .doit(single_tile_kernel,
               d_block_reductions,
               d_out,
-              reduce_grid_size,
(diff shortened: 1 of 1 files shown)


miscco (approved):

elstehle (approved):
Interesting finding 👍  

Seems a bit concerning that device-side `triple_chevron::do_it` isn't type safe. Could imagine that one can easily instantiate a kernel template for different parameter types then are eventually getting passed to the kernel invocation.

Issue #591, opened 2022-11-15
Question: how lookback used in scan avoid loading stale data in L1 cache?


PanZaifeng:
I got confused when reading the code in `agent/single_pass_scan_operators.cuh`.
There is one block that stores the tile data and then updates the status in `SetInclusive` and `SetPartial`:

```cuda
  /**
   * Update the specified tile's partial value and corresponding status
   */
  __device__ __forceinline__ void SetPartial(int tile_idx, T tile_partial)
  {
      // Update tile partial value
      ThreadStore<STORE_CG>(d_tile_partial + TILE_STATUS_PADDING + tile_idx, tile_partial);

      // Fence
      __threadfence();

      // Update tile status
      ThreadStore<STORE_CG>(d_tile_status + TILE_STATUS_PADDING + tile_idx, StatusWord(SCAN_TILE_PARTIAL));
  }
```

Another block waits for the status becomes valid and then loads the tile data in `WaitForValid`:
```cuda
  /**
   * Wait for the corresponding tile to become non-invalid
   */
  __device__ __forceinline__ void WaitForValid(
      int             tile_idx,
      StatusWord      &status,
      T               &value)
  {
      do {
          status = ThreadLoad<LOAD_CG>(d_tile_status + TILE_STATUS_PADDING + tile_idx);

          __threadfence();    // prevent hoisting loads from loop or loads below above this one

      } while (status == SCAN_TILE_INVALID);

      if (status == StatusWord(SCAN_TILE_PARTIAL)) 
          value = ThreadLoad<LOAD_CG>(d_tile_partial + TILE_STATUS_PADDING + tile_idx);
      else
          value = ThreadLoad<LOAD_CG>(d_tile_inclusive + TILE_STATUS_PADDING + tile_idx);
  }
```

I think the code above is unsafe, as the `value` loaded from `d_tile_partial` or `d_tile_inclusive` may be stale. 
As mentioned in [PTX cache operators](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#cache-operators), cache modifiers are only treated as performance hints.
Hence, `value` here can be the one in L1 cache, which may be inconsistent with the L2 cache and global memory, if the two blocks are running on different SMs.

Is it possible for this to happen and cause wrong results?

senior-zero:
Hello, @PanZaifeng! 

[...]

PR #606, opened 2022-12-22
Optimize Decoupled Look-Back

senior-zero:
Decoupled Look-Back is a core of many CUB algorithms. This PR provides a few optimizations that help reduce contention on L2 and improve overall performance. This PR is intended as the first in a series of optimizations/tunings, so selecting the best parameters is out of the scope for now. This PR also addresses the following [issue](https://github.com/NVIDIA/cub/issues/591) by relying on strong operations (`.relaxed`, `.release` etc.) instead of hints (`.cg`).

## Optimizations
- Introduce a larger delay before loading the first look-back window since the data has a low probability of being updated. 
- Try loading tile states before falling into the spin loop. Once the first window is in the partial state, the previous ones should likely be as well. In other words, there's no point in waiting before loading.  
- Introduce a delay into the spin loop of `WaitForValid` to reduce contention and help the signal propagate faster. 
- Make tile state size at least four bytes (used to be two) to distribute the load between a larger number of cache lines. 
- Make flag size `U32` instead of `U8` (for the same reasons above) when message size doesn't let us use a single architectural word, and we have to store flags separately. 

## Fixes
The increase of tile state size revealed issues with `.cg` loads on P100 when the message size doesn't fit the architectural word. The fix consists of voting in the spin loop `while (WARP_ANY((status == SCAN_TILE_INVALID), 0xffffffff));`. Although this might be considered as a breaking change since `ScanTileState<T, false>::WaitForValid` didn't use to be cooperative, I think it's okay since `ScanTileState<T, true>::WaitForValid` is cooperative, and we haven't guaranteed that anyway. 

## Results
To benchmark proposed optimizations, I've selected various GPUs and all algorithms that depend on decoupled look-back. Since the algorithm is sensitive to compute / memory clock ratio, I've run benchmarks with both base and TDP-locked clocks. In general, on large input problem sizes, the speedup is significant enough not to lock clocks. Below is the distribution of execution times for old (main) and new (optimized) decoupled look-back in the case of the device exclusive sum on A100. 

density

Apart from the speedup, the deviation of the new version is smaller. To illustrate broader results, I've grouped multiple [benchmarks](https://github.com/senior-zero/thrust_benchmark/tree/main/benches/cub) by the underlying algorithm. For instance, the `select.if` speedups for different input patterns and operations are combined into a single list as follows:
```
|    T     |  Op  |  Pattern  |  Elements  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |         Diff |   %Diff |  Status  |
|----------|------|-----------|------------|------------|-------------|------------|-------------|--------------|---------|----------|
|   U32    | Mid  |    Seq    |    2^28    |   3.302 ms |       3.31% |   2.484 ms |       0.35% |  -817.584 us | -24.76% |   FAIL   |
|   U32    | Mid  |   Const   |    2^28    |   2.946 ms |       3.06% |   2.148 ms |       0.32% |  -797.100 us | -27.06% |   FAIL   |
|   U32    | Mid  |   Rand    |    2^28    |   3.642 ms |       2.64% |   2.867 ms |       0.35% |  -774.785 us | -21.27% |   FAIL   |
|   U32    | Zero |    Seq    |    2^28    |   3.610 ms |       2.86% |   2.820 ms |       0.35% |  -790.071 us | -21.89% |   FAIL   |
|   U32    | Zero |   Const   |    2^28    |   3.543 ms |       3.05% |   2.750 ms |       0.37% |  -792.352 us | -22.37% |   FAIL   |
|   U32    | Zero |   Rand    |    2^28    |   3.551 ms |       3.04% |   2.748 ms |       0.31% |  -803.270 us | -22.62% |   FAIL   |
|   U32    | Even |    Seq    |    2^28    |   3.358 ms |       3.23% |   2.449 ms |       0.34% |  -908.775 us | -27.07% |   FAIL   |
|   U32    | Even |   Const   |    2^28    |   3.704 ms |       3.05% |   2.748 ms |       0.33% |  -955.250 us | -25.79% |   FAIL   |
|   U32    | Even |   Rand    |    2^28    |   3.389 ms |       2.64% |   2.524 ms |       0.27% |  -864.592 us | -25.51% |   FAIL   |
```
Turns into:
```
speedups = [24.76, 27.06, 21.27, 21.89, 22.37, 22.62, 27.07, 25.79, 25.51]
```
This is further presented as a bar in the bar plot below. Therefore, error bars should be treated as different speedups in different scenarios rather than a run-to-run variance. Here's the aggregate result for base clocks:

base

And the image below is for TDP-locked clocks:

tdp

### DIFF
--- a/cub/agent/single_pass_scan_operators.cuh
+++ b/cub/agent/single_pass_scan_operators.cuh
@@ -230,7 +230,7 @@ struct ScanTileState<T, true>
         int     num_tiles,                          ///< [in] Number of tiles
         size_t  &temp_storage_bytes)                ///< [out] Size in bytes of \t d_temp_storage allocation
     {
-        temp_storage_bytes = (num_tiles + TILE_STATUS_PADDING) * sizeof(TileDescriptor);       // bytes needed for tile status descriptors
+        temp_storage_bytes = (num_tiles + TILE_STATUS_PADDING) * sizeof(TxnWord);       // bytes needed for tile status descriptors
         return cudaSuccess;
     }
 
@@ -595,7 +595,7 @@ struct ReduceByKeyScanTileState<ValueT, KeyT, true>
         int     num_tiles,                          ///< [in] Number of tiles
         size_t  &temp_storage_bytes)                ///< [out] Size in bytes of \t d_temp_storage allocation
     {
-        temp_storage_bytes = (num_tiles + TILE_STATUS_PADDING) * sizeof(TileDescriptor);       // bytes needed for tile status descriptors
+        temp_storage_bytes = (num_tiles + TILE_STATUS_PADDING) * sizeof(TxnWord);       // bytes needed for tile status descriptors
         return cudaSuccess;
     }
 
--- a/cub/agent/single_pass_scan_operators.cuh
+++ b/cub/agent/single_pass_scan_operators.cuh
@@ -36,11 +36,15 @@
 #include <iterator>
 
 #include <cub/config.cuh>
+#include <cub/detail/strong_load.cuh>
+#include <cub/detail/strong_store.cuh>
 #include <cub/detail/uninitialized_copy.cuh>
 #include <cub/thread/thread_load.cuh>
 #include <cub/thread/thread_store.cuh>
 #include <cub/util_device.cuh>
 #include <cub/warp/warp_reduce.cuh>
+
+#include <nv/target>
 
 CUB_NAMESPACE_BEGIN
 
@@ -106,6 +110,44 @@ enum ScanTileStatus
     SCAN_TILE_INCLUSIVE,    // Inclusive tile prefix is available
 };
 
+namespace detail 
+{
+
+template <int Delay, unsigned int GridThreshold = 500>
+__device__ __forceinline__ void delay()
+{
+  NV_IF_TARGET(NV_PROVIDES_SM_70,
+               (if (Delay > 0) 
+                {
+                  if (gridDim.x < GridThreshold) 
+                  {
+                    __threadfence_block();
+                  }
+                  else 
+                  {
+                    __nanosleep(Delay); 
+                  }
+                }));
+}
+
+template <int Delay = 350, unsigned int GridThreshold = 500>
+__device__ __forceinline__ void delay_or_prevent_hoisting()
+{
+  NV_IF_TARGET(NV_PROVIDES_SM_70,
+               (delay<Delay, GridThreshold>();),
+               (__threadfence_block();));
+}
+
+template <int Delay = 350, unsigned int GridThreshold = 500>
+__device__ __forceinline__ void delay_on_dc_gpu_or_prevent_hoisting()
+{
+  NV_DISPATCH_TARGET(
+    NV_IS_EXACTLY_SM_80, (delay<Delay, GridThreshold>();),
+    NV_PROVIDES_SM_70,   (delay<    0, GridThreshold>();),
+    NV_IS_DEVICE,        (__threadfence_block();));
+}
+
+}
 
 /**
  * Tile status interface.
@@ -127,20 +169,20 @@ struct ScanTileState<T, true>
     // Status word type
     using StatusWord = cub::detail::conditional_t<
       sizeof(T) == 8,
-      long long,
+      unsigned long long,
       cub::detail::conditional_t<
         sizeof(T) == 4,
-        int,
-        cub::detail::conditional_t<sizeof(T) == 2, short, char>>>;
+        unsigned int,
+        cub::detail::conditional_t<sizeof(T) == 2, unsigned short, unsigned char>>>;
 
     // Unit word type
     using TxnWord = cub::detail::conditional_t<
       sizeof(T) == 8,
-      longlong2,
+      ulonglong2,
       cub::detail::conditional_t<
         sizeof(T) == 4,
-        int2,
-        cub::detail::conditional_t<sizeof(T) == 2, int, uchar2>>>;
+        uint2,
+        unsigned int>>;
 
     // Device word type
     struct TileDescriptor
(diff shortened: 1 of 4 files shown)


elstehle (commented):
Thanks for the elaborate PR description and experimental evaluation! 

I mostly just have some optional nitpicks. The only question I'd have is related to `delay_on_dc_gpu_or_prevent_hoisting`, which does not `__threadfence_block` for SM 70 and above (with exception for SM 80), while we decided to keep the `__threadfence_block` around in other places.

cub/agent/single_pass_scan_operators.cuh
```diff
@@ -637,11 +682,12 @@ struct ReduceByKeyScanTileState<ValueT, KeyT, true>
 //        value.key   = tile_descriptor.key;
 
         TileDescriptor tile_descriptor;
+
         do
         {
-            __threadfence_block(); // prevent hoisting loads from loop
-            TxnWord alias = ThreadLoad<LOAD_CG>(d_tile_descriptors + TILE_STATUS_PADDING + tile_idx);
-            tile_descriptor = reinterpret_cast<TileDescriptor&>(alias);
+          detail::delay_on_dc_gpu_or_prevent_hoisting();
```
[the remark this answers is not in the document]
senior-zero: That's intentional. Presense of `__threadfence_block()` for pre-Pascal is functional. For Volta+ it's an optimization. Performance evaluation showed that it's better the way it is now, in other words, having a delay for 80 and nothing for Volta+ otherwise for reduce by key operator. For other operators it's not the case.

cub/detail/strong_load.cuh
```diff
@@ -0,0 +1,182 @@
...
+#include <iterator>
+
+CUB_NAMESPACE_BEGIN
+
+#ifndef DOXYGEN_SHOULD_SKIP_THIS // Do not document
+
+namespace detail
+{
+
+static __device__ __forceinline__ uint4 load_relaxed(uint4 const *ptr)
```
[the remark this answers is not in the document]
senior-zero: I don't have preferences and I don't think we have a style guide for that right now. The function is a modified copy of thread load, which also has it this way, so let's keep it for now and rewrite everything once we have settled on a concrete style.

cub/agent/single_pass_scan_operators.cuh
```diff
@@ -106,6 +110,44 @@ enum ScanTileStatus
     SCAN_TILE_INCLUSIVE,    // Inclusive tile prefix is available
 };
 
+namespace detail 
+{
+
+template <int Delay, unsigned int GridThreshold = 500>
+__device__ __forceinline__ void delay()
```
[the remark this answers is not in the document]
senior-zero: Depending on the value, compiler could check only cirtain bits at runtime, so having this as a compile-time parameter might have some implications. I wouldn't like to rebenchmark everything for this change. Maybe once we have a need to use this function with runtime threshold, we'd invest resources to rebenchmark everything.

cub/agent/single_pass_scan_operators.cuh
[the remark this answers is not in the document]
elstehle: Thanks, that's what I wanted to confirm

elstehle (approved):

cub/agent/single_pass_scan_operators.cuh
```diff
@@ -127,20 +169,20 @@ struct ScanTileState<T, true>
     // Status word type
     using StatusWord = cub::detail::conditional_t<
       sizeof(T) == 8,
-      long long,
+      unsigned long long,
```
gonzalobg, line 172: std::uint64_t ?

cub/agent/single_pass_scan_operators.cuh
```diff
@@ -127,20 +169,20 @@ struct ScanTileState<T, true>
...
-        int,
-        cub::detail::conditional_t<sizeof(T) == 2, short, char>>>;
+        unsigned int,
+        cub::detail::conditional_t<sizeof(T) == 2, unsigned short, unsigned char>>>;
 
     // Unit word type
     using TxnWord = cub::detail::conditional_t<
       sizeof(T) == 8,
-      longlong2,
+      ulonglong2,
```
gonzalobg, line 181: std::uint64_t

cub/agent/single_pass_scan_operators.cuh
```diff
@@ -127,20 +169,20 @@ struct ScanTileState<T, true>
     // Status word type
     using StatusWord = cub::detail::conditional_t<
       sizeof(T) == 8,
-      long long,
+      unsigned long long,
       cub::detail::conditional_t<
         sizeof(T) == 4,
-        int,
-        cub::detail::conditional_t<sizeof(T) == 2, short, char>>>;
+        unsigned int,
```
gonzalobg, line 175: std::uint32_t ?

cub/agent/single_pass_scan_operators.cuh
```diff
@@ -127,20 +169,20 @@ struct ScanTileState<T, true>
...
     // Unit word type
     using TxnWord = cub::detail::conditional_t<
       sizeof(T) == 8,
-      longlong2,
+      ulonglong2,
       cub::detail::conditional_t<
         sizeof(T) == 4,
-        int2,
-        cub::detail::conditional_t<sizeof(T) == 2, int, uchar2>>>;
+        uint2,
```
gonzalobg, line 184: std::uint64_t

cub/detail/strong_load.cuh
```diff
@@ -0,0 +1,182 @@
...
+{
+  uint4 retval;
+  NV_IF_TARGET(NV_PROVIDES_SM_70,
+               (asm volatile("ld.relaxed.gpu.v4.u32 {%0, %1, %2, %3}, [%4];"
+                             : "=r"(retval.x), "=r"(retval.y), "=r"(retval.z), "=r"(retval.w)
+                             : _CUB_ASM_PTR_(ptr)
+                             : "memory");),
```
gonzalobg, line 55: I think memory clobber not needed here because load + relaxed does not modify any memory.

gonzalobg (approved):
LGTM

cub/detail/strong_load.cuh
senior-zero: Will be addressed in a separate [issue](https://github.com/NVIDIA/cub/issues/622#issue-1574400395).

cub/agent/single_pass_scan_operators.cuh
senior-zero: Will address in a separate [issue](https://github.com/NVIDIA/cub/issues/623#issue-1574406833)

Issue #52, opened 2016-08-23
Support for custom types in cub::DeviceRadixSort?


maltese:
I couldn't find a better place to ask this question, so pardon me for asking here:

Is there a way to use keys of non-standard types for cub::DeviceRadixSort?

So far I managed to sort arrays of ints, floats, doubles, longs, bools, etc.

But what I would like to sort an array where each item is a custom type:

```
typedef struct {
  char data[32]
} LongLongLong;
```

Unfortunately I get the following error when I try to do that:

`../cub/device/dispatch/dispatch_radix_sort.cuh(254): error: class "cub::Traits<LongLongLong>" has no member "UnsignedBits"
`

Should I maybe add a NumericTraits for my custom type? How would I go about that?

Thank you.

maltese:
What we are basically trying to do is to sort 32-byte long strings. Would it be possible?

egaburov:
Sorting 32 byte strings will take a lot longer than sorting 4 byte ints with radix sort. 8x more bits to sort and 8x more data to move around per pass , netting to 64x increase of work when compared to int 

Comparison sort should fare better in this case. work increase is 8x (modulo comparison operator), because  8x more data will have to be moved around per pass, but the number of passes don't depend on the data type.

Consider thrust::sort for this, which uses merge sort for non primitive data types, but you'll need to provide (optimized) comparison operator. It should work but it won't be fast.

maltese:
Thank you egaburov for the reply. I understand the performance reasoning.

However, an important thing that I forgot to mention is that I would only need to use a 4 byte substring piece of the whole 32 byte string to sort the list. Therefore we do not have 8x more bits to sort, but only 8x more data to move around.

Would this be possible?

Regardless, would you still go with merge sort instead?

Another thing I could think of is to add one pass before the sorting to create a list of just these 4byte substrings along with the indices pointing to the original 32byte string. After that I would scatter the 32byte strings according to the sorted data.

Would appreciate to hear your opinion.

egaburov:
Your proposed solution sounds okay to me, if char[4] can be treated as unsigned int. Use SortPair in this case, with the value in (key,value) pair being either the corresponding index or the whole 32 byte string that will be move around at ever pass. Try both and see which has lower runtime in your case.

daktfi:
Let me bump this a little. Is there any way to sort with key being some custom very wide integer type, like 128 bit or so?
[...]

PR #607, opened 2023-01-02
Support __{u,}int128_t in radix sort

senior-zero:
This PR supports `__int128_t` and `__uint128_t` in block/device radix sort as well as block radix rank. It's partially answering the [question](https://github.com/NVIDIA/cub/issues/52) on 128 bit types support. Performance wise, `__int128_t` utilize BW similarly to 64 bit types.

### DIFF
--- a/cub/util_type.cuh
+++ b/cub/util_type.cuh
@@ -49,12 +49,28 @@
 
 #include <cub/detail/uninitialized_copy.cuh>
 #include <cub/util_arch.cuh>
+#include <cub/util_compiler.cuh>
 #include <cub/util_deprecated.cuh>
 #include <cub/util_macro.cuh>
 #include <cub/util_namespace.cuh>
 
 CUB_NAMESPACE_BEGIN
 
+#ifndef CUB_IS_INT128_ENABLED
+#if defined(__CUDACC_RTC__)
+#if defined(__CUDACC_RTC_INT128__)
+#define CUB_IS_INT128_ENABLED 1
+#endif // !defined(__CUDACC_RTC_INT128__)
+#else  // !defined(__CUDACC_RTC__)
+#if (__CUDACC_VER_MAJOR__ >= 11 || CUDA_VERSION >= 11050) 
+#if (CUB_HOST_COMPILER == CUB_HOST_COMPILER_GCC) || \
+    (CUB_HOST_COMPILER == CUB_HOST_COMPILER_CLANG) || \
+    defined(__ICC) || defined(_NVHPC_CUDA)
+#define CUB_IS_INT128_ENABLED 1
+#endif // GCC || CLANG || ICC || NVHPC
+#endif // CTK >= 11.5
+#endif // !defined(__CUDACC_RTC__)
+#endif // !defined(CUB_IS_INT128_ENABLED)
 
 /**
  * \addtogroup UtilModule
@@ -1185,6 +1201,80 @@ template <> struct NumericTraits<unsigned short> :      BaseTraits<UNSIGNED_INTE
 template <> struct NumericTraits<unsigned int> :        BaseTraits<UNSIGNED_INTEGER, true, false, unsigned int, unsigned int> {};
 template <> struct NumericTraits<unsigned long> :       BaseTraits<UNSIGNED_INTEGER, true, false, unsigned long, unsigned long> {};
 template <> struct NumericTraits<unsigned long long> :  BaseTraits<UNSIGNED_INTEGER, true, false, unsigned long long, unsigned long long> {};
+
+
+#if CUB_IS_INT128_ENABLED 
+template <>
+struct NumericTraits<__uint128_t>
+{
+  using T = __uint128_t;
+  using UnsignedBits = __uint128_t;
+
+  static constexpr Category       CATEGORY    = UNSIGNED_INTEGER;
+  static constexpr UnsignedBits   LOWEST_KEY  = UnsignedBits(0);
+  static constexpr UnsignedBits   MAX_KEY     = UnsignedBits(-1);
+
+  static constexpr bool PRIMITIVE = false;
+  static constexpr bool NULL_TYPE = false;
+
+  static __host__ __device__ __forceinline__ UnsignedBits TwiddleIn(UnsignedBits key)
+  {
+    return key;
+  }
+
+  static __host__ __device__ __forceinline__ UnsignedBits TwiddleOut(UnsignedBits key)
+  {
+    return key;
+  }
+
+  static __host__ __device__ __forceinline__ T Max()
+  {
+    return MAX_KEY;
+  }
+
+  static __host__ __device__ __forceinline__ T Lowest()
+  {
+    return LOWEST_KEY;
+  }
+};
+
+template <>
+struct NumericTraits<__int128_t>
+{
+  using T = __int128_t;
+  using UnsignedBits = __uint128_t;
+
+  static constexpr Category       CATEGORY    = SIGNED_INTEGER;
+  static constexpr UnsignedBits   HIGH_BIT    = UnsignedBits(1) << ((sizeof(UnsignedBits) * 8) - 1);
+  static constexpr UnsignedBits   LOWEST_KEY  = HIGH_BIT;
+  static constexpr UnsignedBits   MAX_KEY     = UnsignedBits(-1) ^ HIGH_BIT;
+
+  static constexpr bool PRIMITIVE = false;
+  static constexpr bool NULL_TYPE = false;
+
+  static __host__ __device__ __forceinline__ UnsignedBits TwiddleIn(UnsignedBits key)
+  {
+    return key ^ HIGH_BIT;
+  };
+
+  static __host__ __device__ __forceinline__ UnsignedBits TwiddleOut(UnsignedBits key)
+  {
+    return key ^ HIGH_BIT;
+  };
+
+  static __host__ __device__ __forceinline__ T Max()
+  {
+    UnsignedBits retval = MAX_KEY;
+    return reinterpret_cast<T&>(retval);
+  }
+
+  static __host__ __device__ __forceinline__ T Lowest()
+  {
+    UnsignedBits retval = LOWEST_KEY;
+    return reinterpret_cast<T&>(retval);
+  }
+};
+#endif
 
 template <> struct NumericTraits<float> :               BaseTraits<FLOATING_POINT, true, false, unsigned int, float> {};
 template <> struct NumericTraits<double> :              BaseTraits<FLOATING_POINT, true, false, unsigned long long, double> {};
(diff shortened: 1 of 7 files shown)


miscco (approved):
That is awesome, thanks a lot. I assume that `int128` implicitly requires `uint128` which has the same switch?

elstehle (approved):
Great work, looks solid!

test/test_device_radix_sort.cu
```diff
@@ -904,9 +904,15 @@ void InitializeSolution(
             // Mask off unwanted portions
             if (num_bits < static_cast<int>(sizeof(KeyT) * 8))
             {
+#if CUB_IS_INT128_ENABLED
+                __uint128_t base = 0;
```
canonizer, line 908: Can `base` be of type `UnsignedBits` corresponding to `KeyT`?

test/test_util.h
```diff
@@ -728,6 +728,43 @@ std::ostream& operator<<(std::ostream& os, const CUB_NS_QUALIFIER::KeyValuePair<
     return os;
 }
 
+#if CUB_IS_INT128_ENABLED
+static std::ostream& operator<<(std::ostream& os, __uint128_t val)
```
canonizer, line 732: Does this have to be defined manually? If yes, there should be tests for it.

test/test_util.h
```diff
@@ -728,6 +728,43 @@ std::ostream& operator<<(std::ostream& os, const CUB_NS_QUALIFIER::KeyValuePair<
...
+{
+  constexpr int max_digits = 40;
```
canonizer, line 734: I'd call it `MAX_DIGITS`, but this is up to you.

test/test_util.h
```diff
@@ -728,6 +728,43 @@ std::ostream& operator<<(std::ostream& os, const CUB_NS_QUALIFIER::KeyValuePair<
...
+  while(val != 0);
+
+  for (; digit != buffer + max_digits; digit++) {
+    os << *digit;
+  }
+
+  return os;
+}
+
+static std::ostream& operator<<(std::ostream& os, __int128_t val)
```
canonizer, line 754: Does this have to be defined manually? If yes, there should be tests for it.

test/test_util.h
```diff
@@ -728,6 +728,43 @@ std::ostream& operator<<(std::ostream& os, const CUB_NS_QUALIFIER::KeyValuePair<
...
+
+  do 
+  {
+    *digit = ascii[val % 10];
+    val /= 10;
+    digit--;
+  }
+  while(val != 0);
+
+  for (; digit != buffer + max_digits; digit++) {
```
canonizer, line 747: I think that at the beginning of this loop, `digit` points to a position where no digit has been written.

test/test_util.h
canonizer, line 732: Also, consider having a separate pull request for these two functions, together with the tests.

canonizer (changes requested):

test/test_util.h
senior-zero: It's a test util only, therefore tests are welcome but not necessery.

test/test_util.h
senior-zero: It's a test util only, therefore tests are welcome but not necessery.

cub/util_type.cuh
```diff
@@ -1186,6 +1202,80 @@ template <> struct NumericTraits<unsigned int> :        BaseTraits<UNSIGNED_INTE
 template <> struct NumericTraits<unsigned long> :       BaseTraits<UNSIGNED_INTEGER, true, false, unsigned long, unsigned long> {};
 template <> struct NumericTraits<unsigned long long> :  BaseTraits<UNSIGNED_INTEGER, true, false, unsigned long long, unsigned long long> {};
 
+
+#if CUB_IS_INT128_ENABLED 
+template <>
+struct NumericTraits<__uint128_t>
```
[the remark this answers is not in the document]
senior-zero: This can't be implemented by inheriting from `BaseTraits` because there's no `BaseTraits` specialization for non-primitive types.

cub/util_type.cuh
```diff
@@ -1186,6 +1202,80 @@ template <> struct NumericTraits<unsigned int> :        BaseTraits<UNSIGNED_INTE
...
+  }
+
+  static __host__ __device__ __forceinline__ T Lowest()
+  {
+    return LOWEST_KEY;
+  }
+};
+
+template <>
+struct NumericTraits<__int128_t>
```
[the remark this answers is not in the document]
senior-zero: Same as above

test/test_device_radix_sort.cu
senior-zero: That's much better, thanks!

test/test_util.h
senior-zero: Nice catch, thanks!

test/catch2_test_printing.cu
```diff
@@ -0,0 +1,34 @@
+#include "test_util.h"
```
canonizer, line 1: Thanks for adding these tests!

test/catch2_test_printing.cu
```diff
@@ -0,0 +1,34 @@
...
+  std::stringstream ss;
+  ss << val;
+  return ss.str();
+}
+
+#if CUB_IS_INT128_ENABLED
+TEST_CASE("Test utils can print __int128", "[test][utils]")
+{
+  REQUIRE( print(__int128_t{0}) == "0" );
+  REQUIRE( print(__int128_t{-1}) == "-1" );
```
canonizer, line 17: Could you also print some strictly positive numbers, and some numbers that don't fit into 64 bits?

test/test_device_radix_sort.cu
```diff
@@ -904,15 +904,11 @@ void InitializeSolution(
...
-                base &= ((__uint128_t{1} << num_bits) - 1) << begin_bit;
-#else 
-                unsigned long long base = 0;
+                using UnsignedBits = typename cub::Traits<KeyT>::UnsignedBits;
+
+                UnsignedBits base = 0;
                 memcpy(&base, &h_keys[i], sizeof(KeyT));
-                base &= ((1ull << num_bits) - 1) << begin_bit;
-#endif
+                base &= ((UnsignedBits{1} << num_bits) - 1) << begin_bit;
```
canonizer, line 911: Will this work if `num_bits == sizeof(KeyT) * 8` and `begin_bit == 0`, e.g. 32 and 0 respectively for `int`?

test/test_block_radix_rank.cu
```diff
@@ -159,9 +159,9 @@ void TestDriver(GenMode gen_mode)
   constexpr int tile_size = BlockThreads * ItemsPerThread;
 
   // Allocate host arrays
-  std::unique_ptr<Key> h_keys(new Key[tile_size]);
-  std::unique_ptr<int> h_ranks(new int[tile_size]);
-  std::unique_ptr<int> h_reference_ranks(new int[tile_size]);
+  std::unique_ptr<Key[]> h_keys(new Key[tile_size]);
```
canonizer, line 162: Not sure if this is related to this pull request, but I'll leave it up to you.

test/test_device_batch_memcpy.cu
```diff
@@ -312,9 +312,9 @@ void RunTest(BufferOffsetT num_buffers,
   using RandomInitAliasT         = uint16_t;
   std::size_t num_aliased_factor = sizeof(RandomInitAliasT) / sizeof(uint8_t);
   std::size_t num_aliased_units  = CUB_QUOTIENT_CEILING(num_total_bytes, num_aliased_factor);
-  std::unique_ptr<uint8_t> h_in(new uint8_t[num_aliased_units * num_aliased_factor]);
-  std::unique_ptr<uint8_t> h_out(new uint8_t[num_total_bytes]);
-  std::unique_ptr<uint8_t> h_gpu_results(new uint8_t[num_total_bytes]);
+  std::unique_ptr<uint8_t[]> h_in(new uint8_t[num_aliased_units * num_aliased_factor]);
+  std::unique_ptr<uint8_t[]> h_out(new uint8_t[num_total_bytes]);
```
canonizer, line 316: Not sure if this is related to this pull request, but I'll leave it up to you.

cub/util_type.cuh
```diff
@@ -1186,6 +1202,80 @@ template <> struct NumericTraits<unsigned int> :        BaseTraits<UNSIGNED_INTE
...
+struct NumericTraits<__uint128_t>
+{
+  using T = __uint128_t;
+  using UnsignedBits = __uint128_t;
+
+  static constexpr Category       CATEGORY    = UNSIGNED_INTEGER;
+  static constexpr UnsignedBits   LOWEST_KEY  = UnsignedBits(0);
+  static constexpr UnsignedBits   MAX_KEY     = UnsignedBits(-1);
+
+  static constexpr bool PRIMITIVE = false;
```
canonizer, line 1217: Is there any difference from `BaseTraits` apart from `PRIMITIVE == false`? What does `PRIMITIVE` mean and what is it used for?

cub/util_type.cuh
```diff
@@ -1186,6 +1202,80 @@ template <> struct NumericTraits<unsigned int> :        BaseTraits<UNSIGNED_INTE
...
+{
+  using T = __int128_t;
+  using UnsignedBits = __uint128_t;
+
+  static constexpr Category       CATEGORY    = SIGNED_INTEGER;
+  static constexpr UnsignedBits   HIGH_BIT    = UnsignedBits(1) << ((sizeof(UnsignedBits) * 8) - 1);
+  static constexpr UnsignedBits   LOWEST_KEY  = HIGH_BIT;
+  static constexpr UnsignedBits   MAX_KEY     = UnsignedBits(-1) ^ HIGH_BIT;
+
+  static constexpr bool PRIMITIVE = false;
```
canonizer, line 1252: Is there any difference from `BaseTraits` apart from `PRIMITIVE == false`? What does `PRIMITIVE` mean and what is it used for?

canonizer (approved):
Approved, provided that the comments are addressed.

test/test_device_radix_sort.cu
senior-zero: It won't, but we never get this case because this code is under the following condition:
```cpp
if (num_bits < static_cast<int>(sizeof(KeyT) * 8))
```

cub/util_type.cuh
senior-zero: `BaseTraits` has many specializations. If the question is related to differences between `BaseTraits<{UN,}SIGNED_INTEGER, true, false, ...>` specializations and `__{u,}int128_t` specializations, there's no difference apart from `PRIMITIVE == false`.

In my understanding, `PRIMITIVE == true` means that the type is fundamental and not part of CUDA extension. For instance, all vector types are marked as not primitive. Since `__{u,}int128_t`is an extension, it falls into the same category. `PRIMITIVE` is also used in the decoupled look-back as a way to distinguish between types that can be combined with a flag to form a single architectural word or not. So, marking `__{u,}int128_t` as primitive would break scan at the moment.

PR #609, opened 2023-01-03
Use libcu++ function objects

senior-zero:
This PR provides a small cleanup and replaces custom implementations of `Equality`, `Inequality`, `Sum`, `Difference`, and `Division` with libcu++ ones for C++14+. Once C++11 support is dropped, we'll remove custom implementations completely.

### DIFF
--- a/cub/thread/thread_operators.cuh
+++ b/cub/thread/thread_operators.cuh
@@ -69,7 +69,7 @@ struct InequalityWrapper
   template <typename T, typename U>
   __host__ __device__ __forceinline__ bool operator()(T &&t, U &&u)
   {
-    return !op(std::forward<T>(t), std::forward<U>(u));
+    return !op(::cuda::std::forward<T>(t), ::cuda::std::forward<U>(u));
   }
 };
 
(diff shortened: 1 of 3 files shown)


cub/thread/thread_operators.cuh
```diff
@@ -51,6 +53,33 @@ CUB_NAMESPACE_BEGIN
...
+  /// Constructor
+  __host__ __device__ __forceinline__ InequalityWrapper(EqualityOp op)
+      : op(op)
+  {}
+
+  /// Boolean inequality operator, returns `t != u`
+  template <typename T, typename U>
+  __host__ __device__ __forceinline__ bool operator()(T &&t, U &&u)
+  {
+    return !op(std::forward<T>(t), std::forward<U>(u));
```
miscco, line 72: I believe that should be 
```suggestion
    return !op(::cuda::std::forward<T>(t), ::cuda::std::forward<U>(u));
```

miscco (approved):
Looks good bar a minor nit

cub/thread/thread_operators.cuh
senior-zero: This PR doesn't introduce any changes into mentioned code.

miscco: I still believe we should use cuda::std functionality instead of plain `std::`

Technically `std::forward` is not annotated with `__device__` so...

senior-zero: @miscco sure! The code originated when we missed libcu++ as a dependency. I meant that std::forward replacement can be done as a different PR, since it isn't related to "Use libcu++ function objects". I've checked and there seems to be not many places, so I can do it as part of this PR.

PR #617, opened 2023-01-19
Fix Clang CUDA

senior-zero:
This PR fixes a few issues that affect Clang CUDA and allows one to build CUB tests with it.

### DIFF
--- a/cub/warp/specializations/warp_reduce_shfl.cuh
+++ b/cub/warp/specializations/warp_reduce_shfl.cuh
@@ -46,6 +46,36 @@
 CUB_NAMESPACE_BEGIN
 
 
+namespace detail 
+{
+
+template <class A = int, class T = A>
+struct reduce_add_exists : std::false_type 
+{};
+
+template <class A>
+struct reduce_add_exists<A, decltype(__reduce_add_sync(0xFFFFFFFF, std::declval<A>()))> : std::true_type 
+{};
+
+template <class A = int, class T = A>
+struct reduce_min_exists : std::false_type 
+{};
+
+template <class A>
+struct reduce_min_exists<A, decltype(__reduce_min_sync(0xFFFFFFFF, std::declval<A>()))> : std::true_type 
+{};
+
+template <class A = int, class T = A>
+struct reduce_max_exists : std::false_type 
+{};
+
+template <class A>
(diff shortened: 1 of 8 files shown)


allisonvacanti (approved):

cub/warp/specializations/warp_reduce_shfl.cuh
```diff
@@ -46,6 +46,36 @@
...
+struct reduce_min_exists<A, decltype(__reduce_min_sync(0xFFFFFFFF, std::declval<A>()))> : std::true_type 
+{};
+
+template <class A = int, class T = A>
+struct reduce_max_exists : std::false_type 
+{};
+
+template <class A>
+struct reduce_max_exists<A, decltype(__reduce_max_sync(0xFFFFFFFF, std::declval<A>()))> : std::true_type 
+{};
```
miscco, line 74: This is a bit overly complicated.

* the second template argument is unneeded
* we do not need to use declval but just `int{}` or even 0 should suffice
* I would prefer to use types from libcu++ so `cuda::std::true_type`
 ```suggestion
template <class = int>
struct reduce_add_exists : cuda::std::false_type 
{};

template <>
struct reduce_add_exists<decltype(__reduce_add_sync(0xFFFFFFFF, 0)) : cuda::std::true_type 
{};

template <class = int>
struct reduce_min_exists : cuda::std::false_type 
{};

template <>
struct reduce_min_exists<decltype(__reduce_min_sync(0xFFFFFFFF, 0))> : cuda::std::true_type 
{};

template <class = int>
struct reduce_max_exists : cuda::std::false_type 
{};

template <>
struct reduce_max_exists<decltype(__reduce_max_sync(0xFFFFFFFF, 0))> : cuda::std::true_type 
{};
```

miscco (approved):
Looks great.

There are some simplifications to be had for the type traits, but nothing major

cub/warp/specializations/warp_reduce_shfl.cuh
senior-zero: > the second template argument is unneeded

We need it to find a specific overload which return type matches the argument type.

> we do not need to use declval but just int{} or even 0 should suffice

The expression has to be dependent for SFINAE to work, it can't be int{}.

> I would prefer to use types from libcu++ so cuda::std::true_type

Agreed, thanks!

PR #618, opened 2023-01-23
Test reduction with bf16 and fp16

senior-zero:
This PR adds device reduce tests for `__half` and `__nv_bfloat16`.


elstehle (approved):
Great work!

PR #620, opened 2023-02-01
Avoid system thrust usage in tests

senior-zero:
I've missed linking Catch2 helper with Thrust specified during CMake configuration step. This led to different Thrust version being used by generator and the tests.


allisonvacanti (approved):
LGTM

Issue #619, opened 2023-01-24
DeviceHistogram::HistogramEven and DeviceHistogram::MultiHistogramEven failing for some cases


Beanavil:
Some change in `cub/device/dispatch/dispatch_histogram.cuh` makes the test `test_hipcub_device_histogram.cu` fail when `HistogramEven` is called with
| CounterT| OffsetT | bins    | lower_level | upper_level |
| ------------| ---------- | -------- | --------------- | ---------------  |
| int           | int         | 65536 | 0                | 65536          |

and

| SampleIteratorT | LevelT |
|-----------------------|-----------|
| int                       | int        |
| double                | float     |

and when `MultiHistogramEven` is called with

| NUM_CHANNELS  | NUM_ACTIVE_CHANNELS |CounterT|OffsetT|bins    |lower_level|upper_level|
| -------------------------- |---------------------------------------|------------|----------|---------|---------------|---------------|
| 1                              | 1                                             |int           |int        |65536|0                |65536        |

and

| SampleIteratorT|LevelT|
|----------------------|---------|
|int                       |int       |
|unsigned short   |int       |
|float                    |int       |
|double                |int       |
|double                |float    |

elstehle:
Thanks for reporting the issue, @Beanavil. The issue has been addressed by https://github.com/NVIDIA/cub/pull/632 and should now be fixed.

PR #632, opened 2023-02-21
Fixes overflows for [Multi]HistogramEven over integral types

elstehle:
This PR addresses issue https://github.com/NVIDIA/cub/issues/619. 

**The issue**
The root cause is our accurate bin computation for integral types, where the product `(sample - min_level) * scale.fraction.bins` may overflow in certain circumstances.
```
return static_cast<int>(((sample - min_level) * scale.fraction.bins) /
                                scale.fraction.range);
```

**The following changes are made:**

When computing histogram over integral types, we perform the bin computation using 32-bit integer arithmetic if both operands each comprise at most two bytes. We use 128-bit integer arithmetic if the common type of samples is a 128-bit type. We use arithmetic on 64-bit types otherwise. If the product `(max_level - min_level) * scale.fraction.bins` would overflow for the used arithmetic type, we return `cudaErrorInvalidValue` on calls to `HistogramEven` and `MultiHistogramEven`.

### DIFF
--- a/cub/device/device_histogram.cuh
+++ b/cub/device/device_histogram.cuh
@@ -72,10 +72,17 @@ struct DeviceHistogram
    *
    * @par
    * - The number of histogram bins is (`num_levels - 1`)
-   * - All bins comprise the same width of sample values: 
-   *   `(upper_level - lower_level) / (num_levels - 1)`
-   * - The ranges `[d_samples, d_samples + num_samples)` and 
-   *   `[d_histogram, d_histogram + num_levels - 1)` shall not overlap 
+   * - All bins comprise the same width of sample values:
+   *   `(upper_level - lower_level) / (num_levels - 1)`.
+   * - If the common type of `SampleT` and `LevelT` is of integral type, the bin for a sample is
+   *   computed as `(sample - lower_level) * (num_levels - 1) / (upper_level - lower_level)`, round
+   *   down to the nearest whole number. To protect against potential overflows, if the product
+   *   `(upper_level - lower_level) * (num_levels - 1)` exceeds the number representable by an
+   *   `uint64_t`, the cuda error `cudaErrorInvalidValue` is returned. If the common type is 128
+   *   bits wide, bin computation will use 128-bit arithmetic and `cudaErrorInvalidValue` will only
+   *   be returned if bin computation would overflow for 128-bit arithmetic.
+   * - The ranges `[d_samples, d_samples + num_samples)` and
+   *   `[d_histogram, d_histogram + num_levels - 1)` shall not overlap
    *   in any way.
    * - `cuda::std::common_type<LevelT, SampleT>` must be valid, and both LevelT
    *   and SampleT must be valid arithmetic types. The common type must be
@@ -245,6 +252,13 @@ struct DeviceHistogram
    * - The number of histogram bins is (`num_levels - 1`)
    * - All bins comprise the same width of sample values:
    *   `(upper_level - lower_level) / (num_levels - 1)`
+   * - If the common type of `SampleT` and `LevelT` is of integral type, the bin for a sample is
+   *   computed as `(sample - lower_level) * (num_levels - 1) / (upper_level - lower_level)`, round
+   *   down to the nearest whole number. To protect against potential overflows, if the product
+   *   `(upper_level - lower_level) * (num_levels - 1)` exceeds the number representable by an
+   *   `uint64_t`, the cuda error `cudaErrorInvalidValue` is returned. If the common type is 128
+   *   bits wide, bin computation will use 128-bit arithmetic and `cudaErrorInvalidValue` will only
+   *   be returned if bin computation would overflow for 128-bit arithmetic.
    * - For a given row `r` in `[0, num_rows)`, let 
    *   `row_begin = d_samples + r * row_stride_bytes / sizeof(SampleT)` and 
    *   `row_end = row_begin + num_row_samples`. The ranges
@@ -434,9 +448,17 @@ struct DeviceHistogram
    * - For channel<sub><em>i</em></sub>, the range of values for all histogram bins
    *   have the same width: 
    *   `(upper_level[i] - lower_level[i]) / (num_levels[i] - 1)`
-   * - For a given channel `c` in `[0, NUM_ACTIVE_CHANNELS)`, the ranges 
-   *   `[d_samples, d_samples + NUM_CHANNELS * num_pixels)` and 
-   *   `[d_histogram[c], d_histogram[c] + num_levels[c] - 1)` shall not overlap 
+   * - If the common type of sample and level is of integral type, the bin for a sample is
+   *   computed as `(sample - lower_level[i]) * (num_levels - 1) / (upper_level[i] -
+   *   lower_level[i])`, round down to the nearest whole number. To protect against potential
+   *   overflows, if, for any channel `i`, the product `(upper_level[i] - lower_level[i]) *
+   *   (num_levels[i] - 1)` exceeds the number representable by an `uint64_t`, the cuda error
+   *   `cudaErrorInvalidValue` is returned. If the common type is 128 bits wide, bin computation
+   *   will use 128-bit arithmetic and `cudaErrorInvalidValue` will only be returned if bin
+   *   computation would overflow for 128-bit arithmetic.
+   * - For a given channel `c` in `[0, NUM_ACTIVE_CHANNELS)`, the ranges
+   *   `[d_samples, d_samples + NUM_CHANNELS * num_pixels)` and
+   *   `[d_histogram[c], d_histogram[c] + num_levels[c] - 1)` shall not overlap
    *   in any way.
    * - `cuda::std::common_type<LevelT, SampleT>` must be valid, and both LevelT
    *   and SampleT must be valid arithmetic types. The common type must be
@@ -633,6 +655,14 @@ struct DeviceHistogram
    * - For channel<sub><em>i</em></sub>, the range of values for all histogram 
    *   bins have the same width: 
    *   `(upper_level[i] - lower_level[i]) / (num_levels[i] - 1)`
+   * - If the common type of sample and level is of integral type, the bin for a sample is
+   *   computed as `(sample - lower_level[i]) * (num_levels - 1) / (upper_level[i] -
+   *   lower_level[i])`, round down to the nearest whole number. To protect against potential
+   *   overflows, if, for any channel `i`, the product `(upper_level[i] - lower_level[i]) *
+   *   (num_levels[i] - 1)` exceeds the number representable by an `uint64_t`, the cuda error
+   *   `cudaErrorInvalidValue` is returned. If the common type is 128 bits wide, bin computation
+   *   will use 128-bit arithmetic and `cudaErrorInvalidValue` will only be returned if bin
+   *   computation would overflow for 128-bit arithmetic.
    * - For a given row `r` in `[0, num_rows)`, and sample `s` in 
    *   `[0, num_row_pixels)`, let 
    *   `row_begin = d_samples + r * row_stride_bytes / sizeof(SampleT)`, 
--- a/cub/device/dispatch/dispatch_histogram.cuh
+++ b/cub/device/dispatch/dispatch_histogram.cuh
@@ -47,6 +47,7 @@
 
 #include <thrust/system/cuda/detail/core/triple_chevron_launch.h>
 
+#include <cuda/std/limits>
 #include <cuda/std/type_traits>
 
 #include <nv/target>
@@ -243,13 +244,45 @@ public:
     struct ScaleTransform
     {
     private:
-      using CommonT = typename cuda::std::common_type<LevelT, SampleT>::type;
-      static_assert(cuda::std::is_convertible<CommonT, int>::value,
+      using CommonT = typename ::cuda::std::common_type<LevelT, SampleT>::type;
+      static_assert(::cuda::std::is_convertible<CommonT, int>::value,
                     "The common type of `LevelT` and `SampleT` must be "
                     "convertible to `int`.");
-      static_assert(cuda::std::is_trivially_copyable<CommonT>::value,
+      static_assert(::cuda::std::is_trivially_copyable<CommonT>::value,
                     "The common type of `LevelT` and `SampleT` must be "
                     "trivially copyable.");
+
+      // An arithmetic type that's used for bin computation of integral types, guaranteed to not
+      // overflow for (max_level - min_level) * scale.fraction.bins. Since we drop invalid samples
+      // of less than min_level, (sample - min_level) is guaranteed to be non-negative. We use the
+      // rule: 2^l * 2^r = 2^(l + r) to determine a sufficiently large type to hold the
+      // multiplication result.
+      // If CommonT used to be a 128-bit wide integral type already, we use CommonT's arithmetic
+      using IntArithmeticT = cub::detail::conditional_t<       //
+        sizeof(SampleT) + sizeof(CommonT) <= sizeof(uint32_t), //
+        uint32_t,                                              //
+#if CUB_IS_INT128_ENABLED
+        cub::detail::conditional_t<                            //
+          (::cuda::std::is_same<CommonT, __int128_t>::value || //
+           ::cuda::std::is_same<CommonT, __uint128_t>::value), //
+          CommonT,                                             //
+          uint64_t>                                            //
+#else
+        uint64_t
+#endif
+        >;
+
+      // Alias template that excludes __[u]int128 from the integral types
+      template <typename T>
+      using is_integral_excl_int128 =
+#if CUB_IS_INT128_ENABLED
+        cub::detail::conditional_t<
+          ::cuda::std::is_same<T, __int128_t>::value && ::cuda::std::is_same<T, __uint128_t>::value,
+          ::cuda::std::false_type,
+          ::cuda::std::is_integral<T>>;
+#else
+        ::cuda::std::is_integral<T>;
+#endif
 
       union ScaleT
       {
@@ -274,7 +307,7 @@ public:
       ScaleT ComputeScale(int num_levels,
                           T   max_level,
                           T   min_level,
-                          cuda::std::true_type /* is_fp */)
+                          ::cuda::std::true_type /* is_fp */)
       {
         ScaleT result;
         result.reciprocal =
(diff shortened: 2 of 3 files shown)


cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -251,6 +251,14 @@ public:
                     "The common type of `LevelT` and `SampleT` must be "
                     "trivially copyable.");
 
+      // An arithmetic type that's used for bin computation of integral SampleT, guaranteed to not
+      // overflow for (max_level - min_level) * scale.fraction.bins. Since we drop invalid samples
+      // of less than min_level, (sample - min_level) is guaranteed to be non-negative. We use the
+      // rule: 2^l * 2^r = 2^(l + r) to determine a sufficiently large type to hold the
+      // multiplication result
+      using IntArithmeticT = cub::detail::
+        conditional_t<(sizeof(SampleT) + sizeof(CommonT) <= sizeof(uint32_t)), uint32_t, uint64_t>;
```
senior-zero, line 260: Optional: something like this might be easier to read:
```suggestion
      using IntArithmeticT = cub::detail::conditional_t<       //
        sizeof(SampleT) + sizeof(CommonT) <= sizeof(uint32_t), //
        uint32_t,                                              //
        uint64_t>;
```

cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -357,14 +365,12 @@ public:
...
-                     T      min_level,
-                     ScaleT scale,
-                     cuda::std::false_type /* is_fp */)
+      __host__ __device__ __forceinline__ int
+      ComputeBin(T sample, T min_level, ScaleT scale, cuda::std::false_type /* is_fp */)
       {
-        return static_cast<int>(((sample - min_level) * scale.fraction.bins) /
-                                scale.fraction.range);
+        return static_cast<int>((static_cast<IntArithmeticT>(sample - min_level) *
+                                 static_cast<IntArithmeticT>(scale.fraction.bins)) /
```
senior-zero, line 372: Would there be any value in storing `fraction.bins` and `fraction.range` in some combination of `CommonT` / `IntArithmeticT` instead of casting? We might reuse / rename `CommonT` to `ComputeT` and conditionally switch between common type and the integer logic.

cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -1069,12 +1113,17 @@ public:
...
 
-                if (num_output_levels[channel] > max_levels)
-                    max_levels = num_output_levels[channel];
+              if (error != cudaSuccess)
+              {
+                return error;
+              }
+
+              if (num_output_levels[channel] > max_levels)
+                max_levels = num_output_levels[channel];
```
senior-zero, line 1126:
```suggestion
              if (num_output_levels[channel] > max_levels)
              {
                max_levels = num_output_levels[channel];
              }
```

cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -1069,12 +1113,17 @@ public:
...
-                privatized_decode_op[channel].Init(num_output_levels[channel],
-                                                   upper_level[channel],
-                                                   lower_level[channel]);
+              cudaError_t error = privatized_decode_op[channel].Init(num_output_levels[channel],
+                                                                     upper_level[channel],
+                                                                     lower_level[channel]);
 
-                if (num_output_levels[channel] > max_levels)
-                    max_levels = num_output_levels[channel];
+              if (error != cudaSuccess)
```
senior-zero, line 1120: Do you think `CubDebug` might be helpful here?

cub/device/device_histogram.cuh
```diff
@@ -72,10 +72,15 @@ struct DeviceHistogram
    *
    * @par
    * - The number of histogram bins is (`num_levels - 1`)
-   * - All bins comprise the same width of sample values: 
-   *   `(upper_level - lower_level) / (num_levels - 1)`
-   * - The ranges `[d_samples, d_samples + num_samples)` and 
-   *   `[d_histogram, d_histogram + num_levels - 1)` shall not overlap 
+   * - All bins comprise the same width of sample values:
+   *   `(upper_level - lower_level) / (num_levels - 1)`.
+   * - For an integral `SampleT` the bin for a sample is computed as `(sample - lower_level) *
```
senior-zero, line 77: We do not define what `SampleT` is, but I guess it should be obvious.

cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -1069,12 +1113,17 @@ public:
...
+              {
+                return error;
```
senior-zero, line 1122: I believe that you return the error earlier than setting `temp_storage_bytes`. If users accidentally ignore return code, they risk allocating temporary storage of a random size. This random size is usually larger than available memory. In any way, the contract is that we set temporary storage in the first invocation of the API, so the users are allowed to avoid initializing it. Let's set `temp_storage_bytes` to something meaningful here just in case (probably `0`).

senior-zero (commented):
Thank you for the great work! A few minor comments below. After they are addressed, I'd like to take another look.

cub/device/dispatch/dispatch_histogram.cuh
elstehle: I've had a similar thought, but had decided against it, because currently `fraction.bins` and `fraction.range` can be packed into a single word with _potentially_ lower register footprint than if we'd store them as `ComputeT`. For `CommonT` of smaller than 32-bit word, those have been promoted anyways. So, the only difference for those types is that, now, we're just explicitly casting those (instead of implicit integer promotion). 

If you feel strongly about it, I'm also ok to change it, though 🙂

cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -357,14 +365,12 @@ public:
       }
 
       template <typename T>
-      __host__ __device__ __forceinline__
-      int ComputeBin(T      sample,
-                     T      min_level,
-                     ScaleT scale,
-                     cuda::std::false_type /* is_fp */)
+      __host__ __device__ __forceinline__ int
+      ComputeBin(T sample, T min_level, ScaleT scale, cuda::std::false_type /* is_fp */)
```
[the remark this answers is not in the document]
elstehle: As per your suggestion, changed logic to use `::cuda::std::is_integral<T>`

cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -392,16 +398,54 @@ public:
...
+                                                           cuda::std::true_type /* is_fp */)
+      {
+        return false;
+      }
+
+      __host__ __device__ __forceinline__ bool MayOverflow(CommonT num_bins,
+                                                           cuda::std::false_type /* is_fp */)
+      {
+        return static_cast<IntArithmeticT>(m_max - m_min) >
+               (Traits<IntArithmeticT>::Max() / static_cast<IntArithmeticT>(num_bins));
```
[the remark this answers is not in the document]
elstehle: Good point about custom types and `Traits`. As per your suggestion above, I've changed the logic to use ::cuda::std::is_integral<T>

cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -392,16 +398,54 @@ public:
...
+      }
+
+#ifdef __CUDA_FP16_TYPES_EXIST__
+      __host__ __device__ __forceinline__ bool MayOverflow(__half /*num_bins*/) { return false; }
```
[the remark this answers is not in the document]
elstehle: This should be resolved by using `::cuda::std::is_integral<T>` now.

cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -1069,12 +1113,17 @@ public:
 
             for (int channel = 0; channel < NUM_ACTIVE_CHANNELS; ++channel)
             {
-                privatized_decode_op[channel].Init(num_output_levels[channel],
-                                                   upper_level[channel],
-                                                   lower_level[channel]);
+              cudaError_t error = privatized_decode_op[channel].Init(num_output_levels[channel],
```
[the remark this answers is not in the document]
elstehle: Refactored this part to only return an error when _not_ querying temporary storage requirements.

cub/device/dispatch/dispatch_histogram.cuh
elstehle: Added `CubDebug`.

cub/device/dispatch/dispatch_histogram.cuh
elstehle: Good point 👍  Wasn't very sure about our guarantees. But I can totally see how users would not check the return value when querying memory requirements.

cub/device/dispatch/dispatch_histogram.cuh
senior-zero: @elstehle I agree with your arguments. Just thought that we might save a few device-side conversions and a simplify the code a bit. I'm not sure about performance implications here, so it's entirely optional suggestion. Feel free to ignore.

cub/device/dispatch/dispatch_histogram.cuh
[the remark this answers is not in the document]
elstehle: Thanks for clarifying. I think what we want is to always return the error in case of an overflow but also make sure to return a proper value for `temp_storage_bytes`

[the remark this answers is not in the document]
elstehle: Changed to return `temp_storage_bytes` of `1` when invoked for a scenario that would overflow 👍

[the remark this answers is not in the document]
elstehle: ...and also added tests to verify we initialize `temp_storage_bytes` 🙂

test/test_device_histogram.cu
```diff
@@ -426,77 +426,72 @@ struct SearchTransform
...
+  {
+    if ((sample < min) || (sample >= max))
+    {
+      // Sample out of range
+      return num_levels;
     }
+
+    // Accurate bin computation following the arithmetic we guarantee in the HistoEven docs
+    return static_cast<int>(
+      (static_cast<uint64_t>(sample - min) * static_cast<uint64_t>(num_levels - 1)) /
```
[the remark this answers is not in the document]
elstehle: Good point! 👍 I've added a specialisation for `half_t` that keeps the current behaviour for bin computation.

cub/device/device_histogram.cuh
elstehle: Strictly speaking, this is `If std::common_type<LevelT, SampleT> is integral, the bin for a sample is computed as ...`. It's more accurate but also a bit more convoluted. What's your preference?

senior-zero: @elstehle good point! I'd spell that, but probably in natural language instead, something like: "If the common type of level and sample is of integral type, ..."

test/test_device_histogram.cu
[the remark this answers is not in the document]
senior-zero: I'd prefer a bit different approach. I'd leave existing version as primary template, while specializing for integral and floating types. What do you think?

```cpp
// primary template
template <typename LevelT, typename = void>
struct ScaleTransform;

// integers
template <typename LevelT>
struct ScaleTransform<LevelT, 
                      typename ::cuda::std::enable_if<::cuda::std::is_integral<LevelT>::value>::type>;

// fp
template <typename LevelT>
struct ScaleTransform<LevelT, 
                      typename ::cuda::std::enable_if<
                            ::cuda::std::is_floating_point<LevelT>::value 
                            #if TEST_HALF_T
                        || ::cuda::std::is_same<LevelT, half_t>::value
                            #endif
                        >::type
                      >;
```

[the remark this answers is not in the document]
elstehle: Adopted your suggestion but also had to change the logic for computing the scale and bins for `half_t`, because `half_t` were previously (i.e., before this PR) covered by the `ScaleTransform` template that was also specialised for integral types.

cub/device/device_histogram.cuh
elstehle: Opted for the more precise formulation. Please have another look and resolve, if you're satisfied 🙂

cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -243,14 +243,24 @@ public:
...
 
+      // An arithmetic type that's used for bin computation of integral SampleT, guaranteed to not
+      // overflow for (max_level - min_level) * scale.fraction.bins. Since we drop invalid samples
+      // of less than min_level, (sample - min_level) is guaranteed to be non-negative. We use the
+      // rule: 2^l * 2^r = 2^(l + r) to determine a sufficiently large type to hold the
+      // multiplication result
+      using IntArithmeticT = cub::detail::conditional_t<       //
+        sizeof(SampleT) + sizeof(CommonT) <= sizeof(uint32_t), //
+        uint32_t,                                              //
+        uint64_t>;
```
senior-zero, line 262: libcu++ recognizes `__{u,}int128_t`:
```cpp
#ifndef _LIBCUDACXX_HAS_NO_INT128
template <>          struct __libcpp_is_integral<__int128_t>         : public true_type {};
template <>          struct __libcpp_is_integral<__uint128_t>        : public true_type {};
#endif
```
Would this code trim the arithmetic value to `uint64`?

cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -1069,12 +1114,24 @@ public:
...
-                    max_levels = num_output_levels[channel];
+              error = privatized_decode_op[channel].Init(num_output_levels[channel],
+                                                         upper_level[channel],
+                                                         lower_level[channel]);
+              if (CubDebug(error != cudaSuccess))
+              {
+                // Make sure to also return a reasonable value for `temp_storage_bytes` in case of
+                // an overflow of the bin computation, in which case a subsequent algorithm
+                // invocation will also fail
+                if(!d_temp_storage){
```
senior-zero, line 1125:
```suggestion
                if(!d_temp_storage)
                {
```

test/test_device_histogram.cu
```diff
@@ -426,77 +430,107 @@ struct SearchTransform
...
```
senior-zero, line 462: I don't think the initial [suggestion](https://github.com/NVIDIA/cub/pull/632/files#r1115268567) is addressed. If you'd like to repurpose the primary template for integral types only, please, add a static assert that `T` is integral. Also, there might be hard-to-find issues when we test histogram with `__{u,}int_t`, please, add a static assert for T size as well.

test/test_device_histogram.cu
```diff
@@ -426,77 +430,107 @@ struct SearchTransform
...
+  int operator()(SampleT sample)
+  {
+    if ((sample < min) || (sample >= max))
+    {
+      // Sample out of range
+      return num_levels;
     }
-};
 
+    return static_cast<int>((static_cast<half_t>(sample) - min) / scale);
```
senior-zero, line 530: Could you elaborate on why we do not apply `float` logic here?

cub/device/dispatch/dispatch_histogram.cuh
```diff
@@ -346,35 +356,45 @@ public:
...
+        return static_cast<int>((static_cast<IntArithmeticT>(sample - min_level) *
+                                 static_cast<IntArithmeticT>(scale.fraction.bins)) /
+                                static_cast<IntArithmeticT>(scale.fraction.range));
+      }
+
+      /**
+       * @brief Bin computation for custom types
+       */
+      template <typename T,
+                typename ::cuda::std::enable_if<!::cuda::std::is_integral<T>::value, int>::type = 0>
```
senior-zero, line 386: `is_integral` already means not fp. The constrain should be on the root function:
```cpp
      template <typename T>
      __host__ __device__ __forceinline__ int
      ComputeBin(T sample, T min_level, ScaleT scale, ::cuda::std::true_type /* is_fp */)
      {
        return static_cast<int>((sample - min_level) * scale.reciprocal);
      }

      /**
       * @brief Bin computation for custom types
       */
      template <typename T>
      __host__ __device__ __forceinline__ int
      ComputeBin(T sample, T min_level, ScaleT scale, ::cuda::std::false_type /* is_fp */)
      {
        return static_cast<int>(((sample - min_level) * scale.fraction.bins) /
                                scale.fraction.range);
      }

      template <typename T, typename ::cuda::std::enable_if<!::cuda::std::is_integral<T>::value, int>::type = 0>
      __host__ __device__ __forceinline__ int ComputeBin(T sample, T min_level, ScaleT scale)
      {
        return this->ComputeBin(sample, min_level, scale, ::cuda::std::is_floating_point<T>{});
      }

      template <typename T, typename ::cuda::std::enable_if<::cuda::std::is_integral<T>::value, int>::type = 0>
      __host__ __device__ __forceinline__ int ComputeBin(T sample, T min_level, ScaleT scale)
      {
        return static_cast<int>((static_cast<IntArithmeticT>(sample - min_level) *
                                 static_cast<IntArithmeticT>(scale.fraction.bins)) /
                                static_cast<IntArithmeticT>(scale.fraction.range));
      }
```

cub/device/dispatch/dispatch_histogram.cuh
elstehle: Done 👍

cub/device/dispatch/dispatch_histogram.cuh
elstehle: Looks a bit cleaner, thanks. Adopted your suggestion 👍

test/test_device_histogram.cu
elstehle: Sorry, I just had to forget to remove the old "dead" code path here. Removed it now.

cub/device/dispatch/dispatch_histogram.cuh
elstehle: Thanks, wasn't aware `::cuda::std::is_integral<T>` considers `__[u]int128_t` as `is_integral`. Per offline discussion, we agreed to: 
- apply promoting to wider integral types only up to `uint64_t`
- use 128-bit wide integer arithmetic only if `CommonT` is already 128 bits wide

senior-zero (approved):
Thank you for the great work! A minor comment below, otherwise, let's squash the commits and I'll start testing as soon as our CI is fixed.

test/test_device_histogram.cu
```diff
@@ -1542,6 +1548,68 @@ void TestIntegerBinCalcs()
   CubDebugExit(g_allocator.DeviceFree(d_samples));
 }
 
+/**
+ * @brief Our bin computation for HistogramEven is guaranteed only for when (max_level - min_level)
+ * * num_bins does not overflow when using uint64_t arithmetic. In case bin computation could
+ * overflow, we expect cudaErrorInvalidValue to be returned.
+ */
+void TestOverflow()
```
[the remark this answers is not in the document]
elstehle: Thanks! That's a good idea. Made it a function template and testing it for various types now. If you're happy with the changes I'll squash the commits.

PR #650, opened 2023-03-10
Force reuse of CUDA arches from thrust.

allisonvacanti:
This only affects the test suite, examples, etc, and will not affect consumers of CUB.

We haven't supported building CUB's test suite as a standalone project for a while, preferring to build Thrust/CUB tests together.

This change prevents CUB from being successful configuring its test suite unless it is inside the Thrust project.

The arch flags in both projectss will be removed soon and replaced with the standard CUDA_ARCHITECTURE CMake properties during the switch to a Thrust/CUB/libcu++ monorepo. This error will be removed when that work is completed.

### DIFF
--- a/cmake/CubCudaConfig.cmake
+++ b/cmake/CubCudaConfig.cmake
@@ -1,132 +1,24 @@
 enable_language(CUDA)
 
-#
-# Architecture options:
-#
-
-set(all_archs 35 37 50 52 53 60 61 62 70 72 75 80 86 90)
-set(arch_message "CUB: Explicitly enabled compute architectures:")
-
-# Thrust sets up the architecture flags in CMAKE_CUDA_FLAGS already. Just
-# reuse them if possible. After we transition to CMake 3.18 CUDA_ARCHITECTURE
-# target properties this will need to be updated.
-if (CUB_IN_THRUST)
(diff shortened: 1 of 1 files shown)


senior-zero (approved):

cmake/CubCudaConfig.cmake
```diff
@@ -1,132 +1,24 @@
...
-
-      string(APPEND arch_message " sm_${arch}")
-      string(APPEND CUB_CUDA_FLAGS_NO_RDC " ${arch_flag}")
-      if (NOT arch IN_LIST no_rdc_archs)
-        string(APPEND CUB_CUDA_FLAGS_RDC " ${arch_flag}")
-      endif()
-    endif()
-  endforeach()
+if (NOT CUB_IN_THRUST)
+  message(FATAL_ERROR
```
robertmaynard, line 4: Since we install this file it means we will break `find_package(cub)` for any consumers

allisonvacanti: No consumers should be including this file other than the Thrust build. It doesn't really need to be installed, and is only used to configure the tests, etc. on dev/ci builds.

robertmaynard (approved):

miscco (approved):

Issue #642, opened 2023-03-03
`cub::DeviceReduce::ArgMin` returns wrong value for INF input


asi1024:
`cub::DeviceReduce::ArgMin` returns a wrong value for an array consisting of only inf. (cc: @leofang @asmeurer)

- reproducer:

```cu
#include <iostream>
#include <cub/cub.cuh>

constexpr int N = 10;
constexpr float INF = std::numeric_limits<float>::infinity();

int main() {
  float* in;
  float* d_in;
  cub::KeyValuePair<int, float>* out;
  cub::KeyValuePair<int, float>* d_out;

  in = new float[N];
  out = new cub::KeyValuePair<int, float>;
  std::fill(in, in + N, INF);

  size_t temp_storage_bytes = 1024;
  void* d_temp_storage;
  cudaMalloc((void**)&d_temp_storage, temp_storage_bytes);

  cudaMalloc((void**)&d_in, sizeof(float) * N);
  cudaMalloc((void**)&d_out, sizeof(float) * N);

  cudaMemcpy(d_in, in, sizeof(float) * N, cudaMemcpyHostToDevice);
  cub::DeviceReduce::ArgMin(d_temp_storage, temp_storage_bytes, d_in, d_out, N);
  cudaMemcpy(out, d_out, sizeof(cub::KeyValuePair<int, float>), cudaMemcpyDeviceToHost);

  std::cout << out->key << std::endl;    // Expected: 0
  std::cout << out->value << std::endl;  // Expected: inf
}
```

- output:

```
1
3.40282e+38
```

Rel. https://github.com/cupy/cupy/issues/7424.

senior-zero:
Hello @asi1024 and thank you for reporting the issue! 

I can reproduce the issue. It's caused by the fact that we are using `std::numeric_limits<T>::max()` as the initial value. This value ends up being compared with `std::numeric_limits<T>::infinity()` as:
```cpp
if ((inf.value < max.value) || ((inf.value == max.value) && (inf.key < max.key))) {
  return b;
}
return a;
``` 
We should probably use `inf` as the initial value.

PR #652, opened 2023-03-11
Fix `cub::DeviceReduce::Arg{Max,Min}` for `inf` values

senior-zero:
This PR addresses the following [issue](https://github.com/NVIDIA/cub/issues/642) related to properly handling `inf` values by `cub::DeviceReduce::Arg{Max,Min}`.

### DIFF
--- a/cub/device/dispatch/dispatch_reduce.cuh
+++ b/cub/device/dispatch/dispatch_reduce.cuh
@@ -51,6 +51,59 @@
 
 CUB_NAMESPACE_BEGIN
 
+namespace detail
+{
+namespace reduce
+{
+
+/**
+ * All cub::DeviceReduce::* algorithms are using the same implementation. Some of them, however,
+ * should use initial value only for empty problems. If this struct is used as initial value with
+ * one of the `DeviceReduce` algorithms, the `init` value wrapped by this struct will only be used
+ * for empty problems; it will not be incorporated into the aggregate of non-empty problems.
+ */
+template <class T>
+struct empty_problem_init_t
+{
+  T init;
+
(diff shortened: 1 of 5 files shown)


cub/device/dispatch/dispatch_reduce.cuh
```diff
@@ -51,6 +51,58 @@
...
+namespace detail
+{
+namespace reduce
+{
+
+/**
+ * All cub::DeviceReduce::* algorithms are using the same implementation. Some of them, however,
+ * should use initial value only for empty problems. This struct is used to distinguish between
+ * mentioned cases.
+ */
```
elstehle, line 63:
```suggestion
/**
 * All cub::DeviceReduce::* algorithms are using the same implementation. Some of them, however,
 * should use initial value only for empty problems. If this struct is used as initial value with
 * one of the `DeviceReduce` algorithms, the `init` value wrapped by this struct will only be used
 * for empty problems; it will not be incorporated into the aggregate of non-empty problems.
 */
```

elstehle (approved):
That's a great solution to the problem. Nice that it doesn't even require touching the `Agent` class.

leofang:
cc: @asi1024 for vis

leofang:
@senior-zero Thanks for quick fix! Question: Would the same bug affect the block-level (or any level blow) reduction? In CuPy they also have CUB block reduction that could kick in in certain code paths.

senior-zero:
Hello @leofang! 

Our `block` and `warp` scope reductions do not provide `Arg{Min,Max}` functionality, so they shouldn't be affected by the bug.

PR #659, opened 2023-03-24
Fix `cub::DeviceSpmv` for empty matrices

senior-zero:
Some inputs are leading to out-of-bounds accesses in `cub::DeviceSpmv`. For instance, if the matrix has many rows and columns but not a single non-zero element, we'll try to access `num_nonzeros - 1`. This PR addresses this issue by short-circuiting into a kernel that fills `y` with `ValueT{0.0}`, which is used a accumulator initializer in other parts of `cub::DeviceSpmv`.


miscco (approved):

miscco:
I am not sure if we want to add tests given that it is essentially deprecated

senior-zero:
@miscco we have a test that leads to the out-of-bounds access without this fix already :)

elstehle (approved):

PR #670, opened 2023-04-04
Add WarpExchangeRegister class

pb-dseifert:
* `WarpExchangeRegister` works like `WarpExchange` semantically, but uses warp shuffles internally instead of shared memory. This pattern is useful in a number of situations:
  * Kernels that only need to exchange a small number of items. When exchanging 2, 4 or 8 items per thread, the number of warp shuffles is 1, 4 and 12, respectively, which is lower than a total of 4, 8 and 16 shared memory transactions, respectively.
  * Kernels that already have a high shared memory budget but still need to exchange data to maintain ILP.
  * Kernels running on GPUs that are shared memory starved, such as Turing or Pascal.

  Inspired by
  * https://forums.developer.nvidia.com/t/transposing-register-held-matrices-with-warp-shuffles-need-help/38652/2
  * https://forums.developer.nvidia.com/t/implement-2d-matrix-transpose-using-warp-shuffle-without-local-memory/208418/11

@senior-zero

### DIFF
--- a/cub/warp/specializations/warp_exchange_shfl.cuh
+++ b/cub/warp/specializations/warp_exchange_shfl.cuh
@@ -1,0 +1,327 @@
+/******************************************************************************
+ * Copyright (c) 2011-2023, NVIDIA CORPORATION.  All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the NVIDIA CORPORATION nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL NVIDIA CORPORATION BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ ******************************************************************************/
+
+#pragma once
+
+#include <cub/config.cuh>
+#include <cub/util_ptx.cuh>
+#include <cub/util_type.cuh>
+
+
+CUB_NAMESPACE_BEGIN
+
+namespace detail
+{
+
+template <typename InputT,
+          int ITEMS_PER_THREAD,
+          int LOGICAL_WARP_THREADS = CUB_PTX_WARP_THREADS>
+class WarpExchangeShfl
+{
+  static_assert(PowerOfTwo<LOGICAL_WARP_THREADS>::VALUE,
+                "LOGICAL_WARP_THREADS must be a power of two");
+
+  static_assert(ITEMS_PER_THREAD == LOGICAL_WARP_THREADS,
+                "WARP_EXCHANGE_SHUFFLE currently only works when ITEMS_PER_THREAD == "
+                "LOGICAL_WARP_THREADS");
+
+  constexpr static bool IS_ARCH_WARP = LOGICAL_WARP_THREADS == CUB_WARP_THREADS(0);
+
+  template <typename OutputT, int IDX>
+  class CompileTimeArrayImpl; // fwd decl for partial specializations
+
+  template <typename OutputT> // terminating partial specialization
+  class CompileTimeArrayImpl<OutputT, ITEMS_PER_THREAD>
+  {
+  protected:
+    // used for dumping back the individual values after transposing
+    InputT (&output_items)[ITEMS_PER_THREAD];
+
+    template <int>
+    __device__ void Foreach(bool, unsigned)
+    {}
+
+  public:
+    __device__ CompileTimeArrayImpl(const InputT (&)[ITEMS_PER_THREAD],
+                                    OutputT (&output_items)[ITEMS_PER_THREAD])
+        : output_items{output_items}
+    {}
+  };
+
+  template <typename OutputT, int IDX> // concrete recursion class
+  class CompileTimeArrayImpl : protected CompileTimeArrayImpl<OutputT, IDX + 1>
+  {
+  protected:
+    InputT val;
+
+    template <int NUM_ENTRIES>
+    __device__ void Foreach(const bool xor_bit_set, const unsigned mask)
+    {
+      // The implementation here is a recursive divide-and-conquer approach
+      // that takes inspiration from:
+      // https://forums.developer.nvidia.com/t/transposing-register-held-matrices-with-warp-shuffles-need-help/38652/2
+      //
+      // At its core, the problem can be boiled down to transposing the matrix
+      //
+      //   A B
+      //   C D
+      //
+      // by swapping the off-diagonal elements/sub-matrices B and C recursively.
+      //
+      // This implementation requires power-of-two matrices. In order to avoid
+      // the use of local or shared memory, all index computation has to occur
+      // at compile-time, since registers cannot be indexed dynamically.
+      // Furthermore, using recursive templates reduces the mental load on the
+      // optimizer, since lowering for-loops into registers oftentimes requires
+      // finagling them with #pragma unroll, which leads to brittle code.
+      //
+      // To illustrate this algorithm, let's pretend we have warpSize = 8,
+      // where t0, ..., t7 denote the 8 threads, and thread i has an array of
+      // size 8 with data = [Ai, Bi, ..., Hi] (the columns in the schematics).
+      //
+      // In the first round, we exchange the largest 4x4 off-diagonal
+      // submatrix. Boxes illustrate the submatrices to be exchanged.
+      //
+      //       ROUND 1
+      //       =======
+      //  t0  t1  t2  t3  t4  t5  t6  t7
+      //                 ┌──────────────┐
+      //  A0  A1  A2  A3 │A4  A5  A6  A7│    NUM_ENTRIES == 4 tells us how many
+      //                 │              │       entries we have in a submatrix,
+      //                 │              │       in this case 4 and the size of
+      //  B0  B1  B2  B3 │B4  B5  B6  B7│       the jumps between submatrices.
+      //                 │              │
+      //                 │              │  1. t[0,1,2,3] data[4] swap with t[4,5,6,7]'s data[0]
+      //  C0  C1  C2  C3 │C4  C5  C6  C7│  2. t[0,1,2,3] data[5] swap with t[4,5,6,7]'s data[1]
+      //                 │              │  3. t[0,1,2,3] data[6] swap with t[4,5,6,7]'s data[2]
+      //                 │              │  4. t[0,1,2,3] data[7] swap with t[4,5,6,7]'s data[3]
+      //  D0  D1  D2  D3 │D4  D5  D6  D7│
+      //                 └──────────────┘
+      // ┌──────────────┐
+      // │E0  E1  E2  E3│ E4  E5  E6  E7
+      // │              │
+      // │              │
+      // │F0  F1  F2  F3│ F4  F5  F6  F7
+      // │              │
+      // │              │
+      // │G0  G1  G2  G3│ G4  G5  G6  G7
+      // │              │
+      // │              │
+      // │H0  H1  H2  H3│ H4  H5  H6  H7
+      // └──────────────┘
+      //
+      //       ROUND 2
+      //       =======
+      //  t0  t1  t2  t3  t4  t5  t6  t7
+      //         ┌──────┐        ┌──────┐
+      //  A0  A1 │A2  A3│ E0  E1 │E2  E3│    NUM_ENTRIES == 2 so we have 2
+      //         │      │        │      │       submatrices per thread and there
+      //         │      │        │      │       are 2 elements between these
+      //  B0  B1 │B2  B3│ F0  F1 │F2  F3│       submatrices.
+      //         └──────┘        └──────┘
+      // ┌──────┐        ┌──────┐          1. t[0,1,4,5] data[2] swap with t[2,3,6,7]'s data[0]
+      // │C0  C1│ C2  C3 │G0  G1│ G2  G3   2. t[0,1,4,5] data[3] swap with t[2,3,6,7]'s data[1]
+      // │      │        │      │          3. t[0,1,4,5] data[6] swap with t[2,3,6,7]'s data[4]
+      // │      │        │      │          4. t[0,1,4,5] data[7] swap with t[2,3,6,7]'s data[5]
+      // │D0  D1│ D2  D3 │H0  H1│ H2  H3
+      // └──────┘        └──────┘
+      //         ┌──────┐        ┌──────┐
+      //  A4  A5 │A6  A7│ E4  E5 │E6  E7│
+      //         │      │        │      │
+      //         │      │        │      │
+      //  B4  B5 │B6  B7│ F4  F5 │F6  F7│
+      //         └──────┘        └──────┘
+      // ┌──────┐        ┌──────┐
+      // │C4  C5│ C6  C7 │G4  G5│ G6  G7
+      // │      │        │      │
+      // │      │        │      │
+      // │D4  D5│ D6  D7 │H4  H5│ H6  H7
+      // └──────┘        └──────┘
+      //
+      //       ROUND 3
+      //       =======
+      //  t0  t1  t2  t3  t4  t5  t6  t7
+      //     ┌──┐    ┌──┐    ┌──┐    ┌──┐
+      //  A0 │A1│ C0 │C1│ E0 │E1│ G0 │G1│    NUM_ENTRIES == 1 so we have 4
+      //     └──┘    └──┘    └──┘    └──┘       submatrices per thread and there
+      // ┌──┐    ┌──┐    ┌──┐    ┌──┐           is 1 element between these
+      // │B0│ B1 │D0│ D1 │F0│ F1 │H0│ H1        submatrices.
+      // └──┘    └──┘    └──┘    └──┘
+      //     ┌──┐    ┌──┐    ┌──┐    ┌──┐  1. t[0,2,4,6] data[1] swap with t[1,3,5,7]'s data[0]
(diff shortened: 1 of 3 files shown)


cub/warp/warp_exchange.cuh
```diff
@@ -468,5 +468,89 @@ public:
...
+    {}
+
+  public:
+    __device__ CompileTimeArrayImpl(InputT (&input_items)[SIZE])
+        : output_items{input_items}
+    {}
+  };
+
+  template <int SIZE, int idx> // concrete recursion class
+  class CompileTimeArrayImpl : protected CompileTimeArrayImpl<SIZE, idx + 1>
```
pb-dseifert, line 499: This is pattern used by [Trove](https://github.com/bryancatanzaro/trove/blob/master/trove/array.h#L36) too and something I've been doing a lot in my code. In my experience, hoisting an array into registers places a major "mental load" on the optimizer, and it often then becomes a game of whack-a-mole by sprinkling `#pragma unroll` in front of various nested loops and praying that the compiler doesn't drop all of into local memory. Using "compile-time" arrays is a lot easier to optimize from the compiler's point of view, since the code never subscripts an array and only accesses values using compile-time indexes.

pb-dseifert:
Register-based vs SMEM-based Transpose:
Screenshot from 2023-04-04 19-54-01
The register-based solution obviously has a slightly higher arithmetic intensity (due to all the `SEL` instructions), but also achieves a much better occupancy, due to only requiring 40(!) registers in order to transpose a 32x32 matrix in a single warp.

cub/warp/warp_exchange.cuh
```diff
@@ -468,5 +468,89 @@ public:
   //@}  end member group
 };
 
+template <typename InputT, int ITEMS_PER_THREAD>
+class WarpExchangeRegister
```
senior-zero, line 472: I'd prefer this to be a specialization instead of a separate API. You can take a look at cub/warp/specializations for reference. Would you have time to refactor your PR to provide something along the following lines?
```cpp
enum WarpExchangeAlgorithm
{
  WARP_EXCHANGE_SHMEM,
  WARP_EXCHANGE_SHUFFLE
};

template <typename InputT,
          int ITEMS_PER_THREAD,
          int LOGICAL_WARP_THREADS  = CUB_PTX_WARP_THREADS,
          int LEGACY_PTX_ARCH       = 0,
          WarpExchangeAlgorithm     = WarpExchangeAlgorithm::WARP_EXCHANGE_SHMEM>
class WarpExchange
// ...
```
This would allow new facility to expose the same API which might facilitate adoption, ease documentation and simplify tests. Regarding the tests, if you manage to introduce the algorithm parameter, would you mind adding the algorithm parameter to the warp exchange [test](`https://github.com/NVIDIA/cub/blob/main/test/catch2_test_warp_exchange.cu`)? You can take a look at the following [example](https://github.com/NVIDIA/cub/blob/6fa6d6e7715d43da8605c7b6766ca51035bdf4ec/test/catch2_test_block_load.cu#L167-L174) for reference.

cub/warp/warp_exchange.cuh
```diff
@@ -468,5 +468,89 @@ public:
...
+{
+  static_assert(PowerOfTwo<ITEMS_PER_THREAD>::VALUE, "ITEMS_PER_THREAD must be a power of two");
+
+  template <int SIZE, int idx>
+  class CompileTimeArrayImpl; // fwd decl for partial specializations
+
+  template <int SIZE> // terminating partial specialization
```
senior-zero, line 479: If `SIZE` is always equal to `ITEMS_PER_THREAD`, we can use `ITEMS_PER_THREAD` directly and remove extra template parameter. This might remove some mental load while reading the code. To workaround partial specialization issue, you could introduce a `DUMMY` parameter, like [here](https://github.com/NVIDIA/cub/blob/6fa6d6e7715d43da8605c7b6766ca51035bdf4ec/cub/block/block_load.cuh#L663).

cub/warp/warp_exchange.cuh
```diff
@@ -468,5 +468,89 @@ public:
   //@}  end member group
 };
 
+template <typename InputT, int ITEMS_PER_THREAD>
+class WarpExchangeRegister
+{
+  static_assert(PowerOfTwo<ITEMS_PER_THREAD>::VALUE, "ITEMS_PER_THREAD must be a power of two");
+
+  template <int SIZE, int idx>
```
senior-zero, line 476: We prefer to capitalize template parameters: 
```suggestion
  template <int SIZE, int IDX>
```

cub/warp/warp_exchange.cuh
```diff
@@ -468,5 +468,89 @@ public:
...
+
+    __device__ ~CompileTimeArrayImpl() { this->output_items[idx] = val; }
+
+    __device__ void Transpose(const unsigned mask) { TransposeImpl<SIZE / 2>(mask); }
+  };
+
+  using CompileTimeArray = CompileTimeArrayImpl<ITEMS_PER_THREAD, 0>;
+
+public:
+  __device__ void Transpose(InputT (&input_items)[ITEMS_PER_THREAD], const unsigned mask = 0xFFFF'FFFFU) const {
```
senior-zero, line 549: Our warp-scope facilities usually support multiple logical warps when you partition the actual warp into smaller contiguous groups of threads. If you refactor the code to be an implementation of `cub::WarpExchange`, you might need to accept lane id and the mask and use those instead of `threadIdx.x`.

cub/warp/warp_exchange.cuh
```diff
@@ -468,5 +468,89 @@ public:
...
+    InputT val;
+
+    template <int NUM_ENTRIES>
+    __device__ void Loop(const bool xor_bit_set, const unsigned mask)
+    {
+      // NOTE: Do *NOT* try to refactor this code to use a reference, since nvcc
+      //       tends to choke on it and then drop everything into local memory.
+      const InputT send_val = (xor_bit_set ? CompileTimeArrayImpl<SIZE, idx>::val
+                                           : CompileTimeArrayImpl<SIZE, idx + NUM_ENTRIES>::val);
+      const InputT recv_val = __shfl_xor_sync(mask, send_val, NUM_ENTRIES, SIZE);
```
senior-zero, line 513: It's optional, but it would be nice if you could document the algorithm logic here.

cub/warp/warp_exchange.cuh
```diff
@@ -468,5 +468,89 @@ public:
...
+      (xor_bit_set ? CompileTimeArrayImpl<SIZE, idx>::val
+                   : CompileTimeArrayImpl<SIZE, idx + NUM_ENTRIES>::val) = recv_val;
+
+      constexpr int next_idx = idx + 1 + ((idx + 1) % NUM_ENTRIES == 0) * NUM_ENTRIES;
+      CompileTimeArrayImpl<SIZE, next_idx>::template Loop<NUM_ENTRIES>(xor_bit_set, mask);
+    }
+
+    // terminate recursion
+    template <int NUM_ENTRIES, std::enable_if_t<!NUM_ENTRIES, int> = 0>
```
senior-zero, line 522: We tend to use `Int2Type` instead of `std::enable_if`. If you prefer `enable_if`, make sure not to use `_t` alias, since we have to support C++11. Also, prefer `::cuda::std`.

senior-zero (commented):
Thank you for your contribution! I've left a few comments below. Please, let me know if you have time to address those. Otherwise, we can take it from here.

cub/warp/warp_exchange.cuh
pb-dseifert: sure! For the general API, how much do you want the (partial) specialization to diverge? For instance, since this purely register based, the `__shared__ typename WarpExchangeT::TempStorage` "callback" doesn't make a lot of sense? Do you want me to omit it? I also don't see how the `ScatterToStriped` APIs can be implemented, since this would likely cause the compiler to not be able to infer the target lanes correctly, again dropping everything to local memory

cub/warp/warp_exchange.cuh
pb-dseifert: I actually tried that, the problem is `nvcc` seems to disallow explicit specialization of template classes in class scope. So I guess the only solution here is to use `DUMMY`?

cub/warp/warp_exchange.cuh
pb-dseifert: sure, I just used lowercase to indicates a subscript index, but I'll change that

cub/warp/warp_exchange.cuh
pb-dseifert: sure, is there a portable way to get the `lane_id` in form of `0 <= lane_id < ITEMS_PER_THREAD`?

cub/warp/warp_exchange.cuh
pb-dseifert: I will document everything once the code is in shape (since I don't want to be editing the Doxygen repeatedly), especially since the underlying idea isn't evident to people that haven't thought a lot about this problem.

cub/warp/warp_exchange.cuh
pb-dseifert: In the [CHANGELOG.md](https://github.com/NVIDIA/cub/blob/main/CHANGELOG.md#deprecated-traits) it says that CUB nowadays targets C++14 and that the STL traits should be used? Do you want me to use the STL traits or `Int2Type`?

cub/warp/warp_exchange.cuh
senior-zero: Please, take a look at cub/warp/specializations/warp_reduce_shfl.cuh for reference. The `WarpExchangeT::TempStorage` would have to be defined as `typedef NullType TempStorage`. Regarding the `ScatterToStriped` make sure to SFINAE it out in the `WarpExchange` if shuffle-based algorithm is required.

cub/warp/warp_exchange.cuh
senior-zero: This is rather C++ problem, explicit specializations are not allowed at class scope. As I suggested above, feel free to use `DUMMY` parameter. It's advantage is that it doesn't have any mental load compared to `SIZE`.

cub/warp/warp_exchange.cuh
senior-zero: `WarpExchange` already have `lane_id` and `mask` computed. Your struct can receive these from `WarpExchange`.

cub/warp/warp_exchange.cuh
senior-zero: I don't anticipate you providing Doxygen documentation as part of this PR. The API is documented for `WarpExchange`. The shuffle algorithm is going to be an implementation detail, so it doesn't have to document API. What I meant is providing some comments regarding the algorithm logic.

cub/warp/warp_exchange.cuh
senior-zero: We deprecated C++11, but the support is not removed yet. I'd prefer `Int2Type` here.

cub/warp/warp_exchange.cuh
pb-dseifert: @senior-zero just for posterity: do you want me to "customize" or partially specialize `class WarpExchange`? The former will require significant refactoring, since there's code like
```
  union _TempStorage
  {
    InputT items_shared[ITEMS_PER_TILE + PADDING_ITEMS];
  }; // union TempStorage

  /// Shared storage reference
  _TempStorage &temp_storage;
```
in `class WarpExchange`, which the warp version wouldn't provide?

senior-zero: > Please, take a look at cub/warp/specializations/warp_reduce_shfl.cuh for reference.

I'd like the same approach as in warp reduce [implementation](https://github.com/NVIDIA/cub/blob/b7e042d43e752ef9b621b24d0322e831f2d162e9/cub/warp/warp_reduce.cuh#L159-L162). The temporary storage will be defined in both cases, it's just that in one of those it'll be an empty struct.

pb-dseifert:
@senior-zero I've amended the code to your suggestions. I haven't documented the impl yet, but will do so soon. I haven't added tests, mainly because there's still one problem:

Let's assume we have an array in global memory of `[0, ..., 63]` where threads load data in a coalesced way into an array of size two, i.e. `int arr[2] = {mem[lane], mem[32 + lane]}` (assuming a threadblock of `(32, 1, 1)`). When setting the logical warp size to 2, I get for the `WARP_EXCHANGE_SHMEM` transpose

```
arr[2] == {0, 1} for lane % 2 == 0
arr[2] == {32, 33} for lane % 2 == 1
```
or
```
 arr[0]:    0   32    0   32    0   32    0   32    0   32    0   32    0   32    0   32    0   32    0   32    0   32    0   32    0   32    0   32    0   32    0   32
 arr[1]:    1   33    1   33    1   33    1   33    1   33    1   33    1   33    1   33    1   33    1   33    1   33    1   33    1   33    1   33    1   33    1   33
```

where columns represent lanes 0 to 31. It is apparent that the data from lanes in [2, 31] gets lost. Switching to `WARP_EXCHANGE_SHUFFLE`, I get after the transpose

```
 arr[0]:    0   32    2   34    4   36    6   38    8   40   10   42   12   44   14   46   16   48   18   50   20   52   22   54   24   56   26   58   28   60   30   62
 arr[1]:    1   33    3   35    5   37    7   39    9   41   11   43   13   45   15   47   17   49   19   51   21   53   23   55   25   57   27   59   29   61   31   63
```
Given how `__shfl_xor_sync` works, I don't see a solution to making them work identically without a broadcast into `[LOGICAL_WARP_THREADS, 31]` after the core transpose procedure. Any ideas?

senior-zero:
> I've amended the code to your suggestions. 

@pb-dseifert Thank you, we are getting closer, but I'm afraid that the scheme I suggested is still not implemented. You've fused two implementations in `WarpExchange`. My suggestion was:

> I'd like the same approach as in warp reduce [implementation](https://github.com/NVIDIA/cub/blob/b7e042d43e752ef9b621b24d0322e831f2d162e9/cub/warp/warp_reduce.cuh#L159-L162).

Could you extract shared memory implementation into `warp/specializations/warp_exchange_shfl.cuh` and the new shuffle implementation into `warp/specializations/warp_exchange_shfl.cuh`? Then the warp exchange would contain:

```cpp
using InternalWarpExchange = 
  cub::detail::conditional_t<
    WARP_EXCHANGE_ALGORITHM == WarpExchangeAlgorithm::WARP_EXCHANGE_SHMEM,
    WarpExchangeShfl<T, ITEMS_PER_THREAD, LOGICAL_WARP_THREADS>,
    WarpExchangeSmem<T, ITEMS_PER_THREAD, LOGICAL_WARP_THREADS>>;
```

I'm sorry for nit picking, but I'd also prefer `WARP_EXCHANGE_SMEM`. 

Regarding your main question, is there any chance that you've passed a single temporary storage to all virtual warps? The contract is that each virtual warp gets it's own temporary storage. In the code below, I get the same output as `WARP_EXCHANGE_SHUFFLE`:

```cpp
constexpr int items_per_thread = 2;
constexpr int warp_size = 32;
constexpr int virtual_warp_size = 2;
constexpr int n = items_per_thread * warp_size;
constexpr int virtual_warps = warp_size / virtual_warp_size;

__global__ void kernel(int *data) {
  using warp_exchange_t = cub::WarpExchange<int, items_per_thread, virtual_warp_size>;
  using temp_storage_t = warp_exchange_t::TempStorage;
  __shared__ temp_storage_t temp_storage[virtual_warps];
  const int warp_id = threadIdx.x / virtual_warp_size;

  int thread_data[items_per_thread];
  for (int i = 0; i < items_per_thread; ++i) {
    thread_data[i] = data[warp_size * i + threadIdx.x];
  }

  warp_exchange_t(temp_storage[warp_id]).StripedToBlocked(thread_data, thread_data);

  for (int i = 0; i < items_per_thread; ++i) {
    data[warp_size * i + threadIdx.x] = thread_data[i];
  }
}

int main() {
  thrust::device_vector<int> data(n);
  thrust::sequence(data.begin(), data.end());

  kernel<<<1, warp_size>>>(thrust::raw_pointer_cast(data.data()));
  cudaDeviceSynchronize();

  thrust::host_vector<int> result = data;
  for (int i = 0; i < items_per_thread; i++) {
    std::cout << "arr[" << i << "] = ";
    for (int thread = 0; thread < warp_size; thread++) {
      std::cout << result[warp_size * i + thread] << " ";
    }
    std::cout << std::endl;
  }
}

// arr[0] = 0 32 2 34 4 36 6 38 8 40 10 42 12 44 14 46 16 48 18 50 20 52 22 54 24 56 26 58 28 60 30 62 
// arr[1] = 1 33 3 35 5 37 7 39 9 41 11 43 13 45 15 47 17 49 19 51 21 53 23 55 25 57 27 59 29 61 31 63 
```

pb-dseifert:
@senior-zero sorry, this took a little longer than anticipated.
Yes, I forgot to use multiple temporary storage arrays for the different logical warps. That made it work.

I've split the actual implementations into different files. I've also documented the recursive approach in the shuffle implementation. I hope this is more to your liking, I went with private inheritance, since it avoids most of the compositional boilerplate in `WarpExchange`.

I haven't extended the testsuite, since I'm still looking into how to exactly invoke it.

cub/warp/warp_exchange.cuh
pb-dseifert: implementation documented

cub/warp/warp_exchange.cuh
```diff
@@ -36,11 +36,30 @@
...
 
 
 CUB_NAMESPACE_BEGIN
 
 
+enum class WarpExchangeAlgorithm
+{
+  WARP_EXCHANGE_SMEM,
+  WARP_EXCHANGE_SHUFFLE,
+};
```
senior-zero, line 50: I'm afraid we'll have to go with `enum WarpExchangeAlgorithm` for now. 

Current usage of other algorithms is `cub::BLOCK_LOAD_X`, having it in the enum class leads to duplication `cub::BlockLoadAlgorithm::BLOCK_LOAD_X`, which can be addressed by removing the duplication in each value: `cub::BlockLoadAlgorithm::DIRECT`. I'd prefer this, but it's not consistent with the rest of the library.

cub/warp/warp_exchange.cuh
```diff
@@ -230,24 +226,7 @@ public:
...
-    }
-    WARP_SYNC(member_mask);
-
-    for (int item = 0; item < ITEMS_PER_THREAD; item++)
-    {
-      const int idx = LOGICAL_WARP_THREADS * item + lane_id;
-      output_items[item] = temp_storage.items_shared[idx];
-    }
-  }
+  using InternalWarpExchange::BlockedToStriped;
```
senior-zero, line 229: I'm afraid we'll have go the duplication route here. When people follow the API, they have to arrive at the function signature. Having them to go into the implementation details is not great. It'd also prevent some potential issues with mismatch of specializations signatures. Could you please implement it in the following form:

```cpp
  Alg(InputT (&items)[ITEMS_PER_THREAD],
      OffsetT (&ranks)[ITEMS_PER_THREAD])
  {
    InternalWarpExchange::Alg(items, ranks);
  }
```

You can omit addressing this comment in this PR if you don't have time. I can file an issue and address this later.

cub/warp/specializations/warp_exchange_shfl.cuh
```diff
@@ -0,0 +1,291 @@
...
+  using TempStorage = NullType;
+
+  WarpExchangeShfl() = delete;
+
+  explicit __device__ __forceinline__ WarpExchangeShfl(TempStorage &)
+      : lane_id(IS_ARCH_WARP ? LaneId() : (LaneId() % LOGICAL_WARP_THREADS))
+      , warp_id(IS_ARCH_WARP ? 0 : (LaneId() / LOGICAL_WARP_THREADS))
+      , member_mask(WarpMask<LOGICAL_WARP_THREADS>(warp_id))
+  {}
+
```
senior-zero, line 274: I think this specialization should provide all the methods. Methods that do not have implementation, should have static assert with dependent false, saying that this specialization is not supporting given use case:
```cpp
template<typename T>
struct dependent_false
{
    static constexpr bool value = false;
};

template <typename OutputT>
__device__ __forceinline__ void X(const InputT (&input_items)[ITEMS_PER_THREAD],
                                  OutputT (&output_items)[ITEMS_PER_THREAD])
{
  static_assert(dependent_false<OutputT>::value, "Shuffle specialization of warp exchange does not support X");
}
```

This would produce user-friendly errors in case of the API misuse. You can omit addressing this comment in this PR if you don't have time. I can file an issue and address this later.

senior-zero (commented):
This is excellent work. Thank you! The comment describing the shuffle exchange is incredible, and the specialization scheme is consistent with the rest of the library. 

I've left a few minor comments that must be addressed before merging since they affect API and would be hard to fix without the deprecation procedure. I've also left a few optional comments. You can defer fixing those to us. 

Most importantly, we are moving this repository, so I can't merge your PR here. Could you please reopen it against the following repo https://github.com/NVIDIA/cccl? Once you do that, explaining how to run cub tests on your end will be much easier.

pb-dseifert:
@senior-zero I've addressed all your comments, could you check again? Once you're happy here, I'll open a fresh PR in `cccl`

pb-dseifert:
https://github.com/NVIDIA/cccl/pull/256

PR #671, opened 2023-04-06
Support user-defined types in radix sort and introduce initial sphinx docs

senior-zero:
This PR introduces support for user-defined types in the block / device radix sort (addresses https://github.com/NVIDIA/cub/issues/52):

```cpp
struct custom_t
{
  std::uint16_t i;
  int unused;
  float f;
};

struct decomposer_t
{
  __host__ __device__ //
  ::cuda::std::tuple<std::uint16_t&, float&> operator()(custom_t &key) const
  {
    return {key.i, key.f};
  }
};

cub::DeviceRadixSort::SortKeys(d_temp_storage,
                               temp_storage_bytes,
                               d_in,
                               d_out,
                               num_items,
                               decomposer_t{});
```

There's no difference in performance between sorting a key that's a pair of `int32` and sorting a single `int64` key. The PR also doesn't affect SASS for fundamental types. 

Documentation for the new API is written using sphinx, whose initial support is also introduced by this PR. Doxygen is temporarily broken, and a follow-up PR will provide the complete transition to sphinx. To build sphinx docs:

```bash
:cd docs 
:./gen_docs.sh 
:open _build/docs/CUB/latest/index.html
```


docs/index.rst
```diff
@@ -0,0 +1,497 @@
...
+
+What is CUB?
+==================================================
+
+CUB provides state-of-the-art, reusable software components for every layer 
+of the CUDA programming model:
+
+* **Parallel primitives**
+
+  * **Warp-wide "collective" primitives**
```
jrhemstad, line 23: Do Doxygen groups no longer work with the Sphinx infrastructure? 

I'm comparing the new docs to the old docs page, and these used to link to the Warp/Block/Device modules from Doxygen, e.g., https://nvlabs.github.io/cub/group___warp_module.html

senior-zero: I haven't found how to setup automatic grouping yet.

cub/agent/agent_radix_sort_downsweep.cuh
```diff
@@ -193,34 +193,42 @@ struct AgentRadixSortDownsweep
...
+    OffsetT bin_offset[BINS_TRACKED_PER_THREAD];
 
-    // Digit extractor
-    DigitExtractorT digit_extractor;
+    std::uint32_t current_bit; 
+    std::uint32_t num_bits;
 
     // Whether to short-cirucit
-    int             short_circuit;
+    int short_circuit;
```
miscco, line 208: Why is this an `int `rather than a `bool`?

cub/block/block_radix_sort.cuh
```diff
@@ -522,6 +574,135 @@ public:
...
+    //!   The least-significant bit index (inclusive) needed for 
+    //!   key comparison
+    //!
+    //! @param[in] end_bit 
+    //!   The most-significant bit index (exclusive) needed for key 
+    //!   comparison (e.g., `(sizeof(float) + sizeof(long long int)) * 8`)
+    template <class DecomposerT>
+    __device__ __forceinline__         //
+      typename ::cuda::std::enable_if< //
+        !::cuda::std::is_convertible<DecomposerT, int>::value>::type
```
miscco, line 636:
```suggestion
      ::cuda::std::__enable_if_t<!::cuda::std::is_convertible<DecomposerT, int>::value>>
```

Feel free to disregard if you want to avoid `__ugly` constructs

cub/block/radix_rank_sort_operations.cuh
```diff
@@ -130,16 +119,518 @@ struct ShiftDigitExtractor : BaseDigitExtractor<KeyT>
...
+  static constexpr bool value = true;
+};
+
+template <bool... Bs>
+using all_t =            //
+  ::cuda::std::is_same<  //
+    logic_helper_t<Bs...>, //
+    logic_helper_t<true_t<Bs>::value...>>;
+
+struct fundamental_decomposer_t
```
miscco, line 157: I am wondering whether `identity_decomposer_t` would be better, but I am really not sure whether that is a better name

cub/block/radix_rank_sort_operations.cuh
```diff
@@ -130,16 +119,518 @@ struct ShiftDigitExtractor : BaseDigitExtractor<KeyT>
...
+struct is_fundamental_type
+{
+  static constexpr bool value = false;
+};
+
+template <class T>
+struct is_fundamental_type<T, void_t<typename Traits<T>::UnsignedBits>>
+{
+  static constexpr bool value = true;
+};
```
miscco, line 210: Should we use cuda::std:: traits here?

cub/block/radix_rank_sort_operations.cuh
```diff
@@ -130,16 +119,518 @@ struct ShiftDigitExtractor : BaseDigitExtractor<KeyT>
...
+      src_bit_start -= src_size;
     }
+    else
+    {
+      using traits           = traits_t<typename ::cuda::std::remove_cv<T>::type>;
+      using bit_ordered_type = typename traits::bit_ordered_type;
+
+      const std::uint32_t bits_to_copy = min(src_size - src_bit_start, num_bits);
+
+      if (bits_to_copy)
```
miscco, line 462: Should we early return here?

cub/block/radix_rank_sort_operations.cuh
```diff
@@ -130,16 +119,518 @@ struct ShiftDigitExtractor : BaseDigitExtractor<KeyT>
...
+
+  template <class T>
+  __host__ __device__ void operator()(T &src)
+  {
+    const std::uint32_t src_size = sizeof(T) * 8;
+
+    if (src_bit_start >= src_size)
     {
-        return uint32_t(this->ProcessFloatMinusZero(key) >> UnsignedBits(bit_start)) & mask;
+      src_bit_start -= src_size;
```
miscco, line 453: Should we early return here?

cub/block/radix_rank_sort_operations.cuh
```diff
@@ -130,16 +119,518 @@ struct ShiftDigitExtractor : BaseDigitExtractor<KeyT>
...
+
+  template <class FundamentalExtractorT, class DecomposerT>
+  using digit_extractor_t = custom_digit_extractor_t<DecomposerT>;
+
+  template <class DecomposerT>
+  static __host__ __device__ bit_ordered_type min_raw_binary_key(DecomposerT decomposer)
+  {
+    T val{};
+    detail::radix::min_raw_binary_key(decomposer, val);
+    return val;
```
miscco, line 558: I really dislike that pattern here. I would rather write 
```suggestion
    return detail::radix::min_raw_binary_key<T>(decomposer);
```
And pass the type explicitly rather than constructing using and returning

cub/device/dispatch/dispatch_radix_sort.cuh
```diff
@@ -70,11 +70,12 @@ CUB_NAMESPACE_BEGIN
...
-    bool                    ALT_DIGIT_BITS,                 ///< Whether or not to use the alternate (lower-bits) policy
-    bool                    IS_DESCENDING,                  ///< Whether or not the sorted-order is high-to-low
-    typename                KeyT,                           ///< Key type
-    typename                OffsetT>                        ///< Signed integer type for global offsets
+    typename ChainedPolicyT,                 ///< Chained tuning policy
+    bool     ALT_DIGIT_BITS,                 ///< Whether or not to use the alternate (lower-bits) policy
+    bool     IS_DESCENDING,                  ///< Whether or not the sorted-order is high-to-low
+    typename KeyT,                           ///< Key type
+    typename OffsetT,                        ///< Signed integer type for global offsets
+    typename DecomposerT = detail::fundamental_decomposer_t>
```
senior-zero, line 78: Add documentation stating that `DecomposerT` is considered as part of implementation detail and can be changed without bumping the major version.

cub/block/radix_rank_sort_operations.cuh
```diff
@@ -130,16 +119,518 @@ struct ShiftDigitExtractor : BaseDigitExtractor<KeyT>
...
+{
+  static_assert(sizeof...(Ts), "Empty aggregates are not supported");
+
+  // Most radix operations are indifferent to the order of operations.
+  // Conversely, the digit extractor traverses fields from the least significant 
+  // to the most significant to imitate bitset printing where higher bits are on 
+  // the left. It also maps to intuition, where something coming first is more 
+  // important. Therefore, we traverse fields on the opposite order.
+  for_each_member_impl_helper(f, tpl, THRUST_NS_QUALIFIER::make_reversed_index_sequence<sizeof...(Ts)>{});
+}
```
senior-zero, line 186: Compare compilation time performance of fold expressions to see if there's a value in conditionally using them here.

elstehle (commented):
Really solid work! Like how we can keep most of the implementation details unchanged. 

Partial review; review of `BlockRadixRank` and tests remaining. Also want to push `radix_rank_sort_operations` through its paces a bit.

docs/tools/packman/bootstrap/generate_temp_file_name.ps1
```diff
@@ -0,0 +1,161 @@
...
+# CyqGSIb3DQEJEAEEMBwGCSqGSIb3DQEJBTEPFw0yMTA0MDgwOTExMDlaMCsGCyqG
+# SIb3DQEJEAIMMRwwGjAYMBYEFOHXgqjhkb7va8oWkbWqtJSmJJvzMC8GCSqGSIb3
+# DQEJBDEiBCCHEAmNNj2zWjWYRfEi4FgzZvrI16kv/U2b9b3oHw6UVDANBgkqhkiG
+# 9w0BAQEFAASCAQCdefEKh6Qmwx7xGCkrYi/A+/Cla6LdnYJp38eMs3fqTTvjhyDw
+# HffXrwdqWy5/fgW3o3qJXqa5o7hLxYIoWSULOCpJRGdt+w7XKPAbZqHrN9elAhWJ
+# vpBTCEaj7dVxr1Ka4NsoPSYe0eidDBmmvGvp02J4Z1j8+ImQPKN6Hv/L8Ixaxe7V
+# mH4VtXIiBK8xXdi4wzO+A+qLtHEJXz3Gw8Bp3BNtlDGIUkIhVTM3Q1xcSEqhOLqo
+# PGdwCw9acxdXNWWPjOJkNH656Bvmkml+0p6MTGIeG4JCeRh1Wpqm1ZGSoEcXNaof
+# wOgj48YzI+dNqBD9i7RSWCqJr2ygYKRTxnuU
+# SIG # End signature block
```
ericniebler, line 161: What are these signatures? They're nothing that needs to be kept private, are they?

senior-zero: Seems to be public already in the original [repo](https://github.com/NVIDIA-Omniverse/kit-app-template/blob/main/tools/packman/bootstrap/generate_temp_file_name.ps1).

jrhemstad: I also double checked with the packman maintainers and they confirmed that this is just a signature hash that verifies the authenticity of the script. No secrets being leaked here.

test/test_device_radix_sort.cu
```diff
@@ -1952,6 +1952,182 @@ void TestUnspecifiedRanges()
 }
 #endif
 
+#if TEST_KEY_BYTES == 4
+// Following tests check that new decomposer API doesn't break old API. 
+// It's disabled because some compilers don't like implicit conversions, which
+// is required for the test. Once we figure out how to temporarely enable conversion, we can
```
elstehle, line 1958: nit
```suggestion
// is required for the test. Once we figure out how to temporarily enable conversion, we can
```

test/catch2_test_block_radix_sort.cuh
```diff
@@ -0,0 +1,464 @@
...
+  std::transform(h_permutation.begin(),
+                 h_permutation.end(),
+                 result.first.begin(),
+                 [&](std::size_t i) { return h_keys[i]; });
+
+  thrust::host_vector<ValueT> h_values(d_values);
+  std::transform(h_permutation.begin(),
+                 h_permutation.end(),
+                 result.second.begin(),
+                 [&](std::size_t i) { return h_values[i]; });
```
elstehle, line 461: nit: `thrust::gather` with two `thrust::zip` iterators?

test/catch2_test_block_radix_sort.cuh
```diff
@@ -0,0 +1,464 @@
...
+{
+  thrust::host_vector<KeyT> h_keys(d_keys);
+  thrust::host_vector<std::size_t> h_permutation =
+    get_permutation(h_keys, is_descending, begin_bit, end_bit);
+
+  thrust::host_vector<KeyT> result(d_keys.size());
+  std::transform(h_permutation.begin(),
+                 h_permutation.end(),
+                 result.begin(),
+                 [&](std::size_t i) { return h_keys[i]; });
```
elstehle, line 432: nit: `thrust::gather`?

test/catch2_test_block_radix_sort_custom.cu
```diff
@@ -0,0 +1,1060 @@
+/******************************************************************************
+ * Copyright (c) 2011-2022, NVIDIA CORPORATION.  All rights reserved.
```
elstehle, line 2:
```suggestion
 * Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```

test/catch2_test_block_radix_sort.cuh
```diff
@@ -0,0 +1,464 @@
+/******************************************************************************
+ * Copyright (c) 2011-2022, NVIDIA CORPORATION.  All rights reserved.
```
elstehle, line 2: Not sure if this gets 2023 or 2011-2023 🤷 
```suggestion
 * Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```

test/catch2_test_radix_operations.cu
```diff
@@ -0,0 +1,659 @@
+#include <cub/block/radix_rank_sort_operations.cuh>
```
elstehle, line 1: I believe this should also contain our copyright header?

test/catch2_test_block_radix_sort_custom.cu
```diff
@@ -0,0 +1,1060 @@
...
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ ******************************************************************************/
+
+#undef NDEBUG
```
elstehle, line 28: Is this a remainder from testing?
```suggestion
```

elstehle (commented):
Only a few, mostly minor, comments remaining 🙂

elstehle (approved):

test/catch2_test_block_radix_sort_custom.cu
senior-zero: This test relies on asserts in kernels. Assertion doesn't work in release builds. This line (and the one below) enables assertion in release builds.

cub/agent/agent_radix_sort_upsweep.cuh
```diff
@@ -196,30 +200,34 @@ struct AgentRadixSortUpsweep
...
-    __device__ __forceinline__ void Bucket(UnsignedBits key)
+    __device__ __forceinline__ void Bucket(bit_ordered_type key)
     {
         // Perform transform op
-        UnsignedBits converted_key = Traits<KeyT>::TwiddleIn(key);
+        bit_ordered_type converted_key = bit_ordered_conversion::to_bit_ordered(decomposer, key);
 
         // Extract current digit bits
-        UnsignedBits digit = digit_extractor.Digit(converted_key);
+        std::uint32_t digit = digit_extractor().Digit(converted_key);
```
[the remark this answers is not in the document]
senior-zero: I haven't noticed the difference

examples/device/example_device_radix_sort_custom.cu
```diff
@@ -0,0 +1,267 @@
...
+  d_temp_storage = thrust::raw_pointer_cast(temp_storage.data());
+
+  // 3) Sort keys
+  cub::DeviceRadixSort::SortKeys(d_temp_storage,
+                                 temp_storage_bytes,
+                                 d_in,
+                                 d_out,
+                                 num_items,
+                                 decomposer_t{});
+  cudaDeviceSynchronize();
```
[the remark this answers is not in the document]
senior-zero: > This should be superfluous in the absence of a stream parameter, right? 
Absence of a stream doesn't make the API synchronous. Synchronization here was indented to highlight this fact.

cub/block/radix_rank_sort_operations.cuh
senior-zero: Compared with:
```cpp
(f(::cuda::std::get<Is>(tpl)), ...);
```
No difference in SASS, compilation time match current implementation.

cub/block/radix_rank_sort_operations.cuh
senior-zero: The pattern will not disappear if we fix this place, it'll just go down the stack. Since it's an implementation detail that's consistent with the other functions here, I don't see the reason to change it.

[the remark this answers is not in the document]
senior-zero: The pattern will not disappear if we fix this place, it'll just go down the stack. Since it's an implementation detail that's consistent with the other functions here, I don't see the reason to change it.

cub/block/radix_rank_sort_operations.cuh
senior-zero: cuda::std:: traits do not take into account extended floating point types

cub/agent/agent_radix_sort_downsweep.cuh
senior-zero: I don't know why author preferred int here. I also don't think we should spend time verifying that making it bool doesn't affect perf as part of this PR.

cub/block/radix_rank_sort_operations.cuh
senior-zero: I think `identity_decomposer_t` is better. I'll rename it, thanks!

cub/block/radix_rank_sort_operations.cuh
senior-zero: I'm not sure what we'll get from this.

cub/block/radix_rank_sort_operations.cuh
senior-zero: I don't know what we'll get from this.

cub/device/device_radix_sort.cuh
```diff
@@ -247,6 +325,9 @@ struct DeviceRadixSort
     // Unsigned integer type for global offsets.
     using OffsetT = typename detail::ChooseOffsetT<NumItemsT>::Type;
 
+    // TODO API that doesn't accept decomposer should also contain a static
+    //      assert that the key type is fundamental.
+
```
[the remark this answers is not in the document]
senior-zero: I'll file a separate issue to address this later.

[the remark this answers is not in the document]
senior-zero: https://github.com/NVIDIA/cub/issues/693

PR #678, opened 2023-04-19
Fixes non_void_value_t for ptr-to-void iterator types

elstehle:
Currently `cub::detail::non_void_value_t<void*, my_fallback_type>` fails to compile with the error message `reference to void is not allowed` due to instantiating the `iterator_traits` template with a reference to void member type.

This PR fixes the compilation error.

Thanks to @senior-zero for helping to iterate on the fix.

### DIFF
--- a/cub/util_type.cuh
+++ b/cub/util_type.cuh
@@ -37,7 +37,6 @@
 #include <iostream>
 #include <iterator>
 #include <limits>
-#include <type_traits>
 
 #include <cuda.h>
 
@@ -54,6 +53,8 @@
 #include <cub/util_deprecated.cuh>
 #include <cub/util_macro.cuh>
 #include <cub/util_namespace.cuh>
+
+#include <cuda/std/type_traits>
 
 CUB_NAMESPACE_BEGIN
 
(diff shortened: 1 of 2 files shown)


miscco (approved):

cub/util_type.cuh
```diff
@@ -96,19 +96,32 @@ using conditional_t = typename std::conditional<Test, T1, T2>::type;
 template <typename Iterator>
 using value_t = typename std::iterator_traits<Iterator>::value_type;
 
+template <typename It,
+          typename FallbackT,
+          bool = std::is_same<typename std::remove_cv<typename std::remove_pointer<It>::type>::type,
```
senior-zero, line 101: Is there any particular reason not to use `::cuda::std::` instead of `std::` alternatives?

test/catch2_test_util_type.cu
```diff
@@ -0,0 +1,65 @@
...
+#include "catch2_test_helper.h"
+
+CUB_TEST("Tests non_void_value_t", "[util][type]")
+{
+  using fallback_t        = float;
+  using void_fancy_it     = cub::DiscardOutputIterator<std::size_t>;
+  using non_void_fancy_it = cub::CountingInputIterator<int>;
+
+  // falls back for const void*
+  REQUIRE(std::is_same<fallback_t, //
```
senior-zero, line 45: Do you think we could use `STATIC_REQUIRE` instead?

senior-zero (approved):

cub/util_type.cuh
elstehle: No specific reason. Let's stick consistently to `::cuda::std`. Done 👍

test/catch2_test_util_type.cu
elstehle: Thanks! Done 👍

PR #687, opened 2023-05-05
Fixes narrowing conversion warning in tests

elstehle:
To fix the following warning:

```
thrust/thrust/cmake/../..\thrust/detail/vector_base.inl(233): warning C4244: 'argument': conversion from 'int' to 'const char', possible loss of data 

...

thrust/thrust/cmake/../..\thrust/detail/vector_base.inl(233): warning C4244: 'argument': conversion from 'int' to 'const char', possible loss of data 
```


miscco (approved):
I love that it is also much more efficient

Issue #689, opened 2023-05-08
Add policy parameter to allow tuning


senior-zero:
Not all CUB algorithm follow the scheme of accepting tuning policy at dispatch layer. This prevents these algorithms from being tuned.

```[tasklist]
### Dispatch layers to be modified
- [x] Add `SelectedPolicy ` to `DispatchHistogram`
- [x] Add `SelectedPolicy ` to `DispatchReduceByKey`
- [x] Add `SelectedPolicy` to `DeviceRleDispatch`
- [x] Add `SelectedPolicy ` to `DispatchThreeWayPartitionIf`
```

PR #694, opened 2023-05-22
Add policy parameter to allow tuning

senior-zero:
This PR addresses https://github.com/NVIDIA/cub/issues/689. It adds tuning policy to histogram, rle, three way partition, and reduce by key. This breaks kernels, dispatch constructors, along with some member functions taking kernels as arguments. The change is required to tune mentioned algorithms. Furthermore, we consider API taking kernels and the kernels themselves an implementation detail. The dispatch layer should be used as `DispatchAlgorithmT::Dispatch(...)`. The member functions taking kernels are not considered entry points. Therefore, we do not treat changes in this PR as breaking ones. 

Since the tuning policy was just introduced for tuning purposes, the content of tuning policy is subject to change. Newly introduced tuning policies are documented as an implementation detail for now. 

Apart from that, this PR introduces benchmarks for mentioned algorithms and adds some missing tunings for radix sort / H100.

### DIFF
--- a/benchmarks/bench/histogram/multi/range.cu
+++ b/benchmarks/bench/histogram/multi/range.cu
@@ -1,7 +1,6 @@
-#include <cub/device/device_histogram.cuh>
-
 #include <thrust/sequence.h>
 
+#include "../histogram_common.cuh"
 #include <nvbench_helper.cuh>
 
 // %RANGE% TUNE_ITEMS ipt 7:24:1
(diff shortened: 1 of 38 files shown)


benchmarks/bench/histogram/multi/range.cu
```diff
@@ -0,0 +1,198 @@
...
+      return static_cast<SampleT>(std::min(bins, static_cast<OffsetT>(max_key)));
+    }
+    else
+    {
+      return static_cast<SampleT>(bins);
+    }
+  }
+
+  return static_cast<SampleT>(elements);
+}
```
miscco, line 77: Nit: Do we have a common header where we can pull this into?

Maybe `nvbench_helper.cuh`

benchmarks/bench/reduce/by_key.cu
```diff
@@ -0,0 +1,165 @@
...
+                                                           TUNE_LOAD_ALGORITHM,
+                                                           TUNE_LOAD_MODIFIER,
+                                                           cub::BLOCK_SCAN_WARP_SCANS>;
+  };
+
+  using MaxPolicy = Policy350;
+};
+#endif // !TUNE_BASE
+
+#include <cub/device/device_reduce.cuh>
```
miscco, line 38: Is it intentional that the include comes after the stuff above?

If so we should add a comment

benchmarks/bench/run_length_encode/encode.cu
```diff
@@ -0,0 +1,151 @@
...
+                                                           TUNE_LOAD_ALGORITHM,
+                                                           TUNE_LOAD_MODIFIER,
+                                                           cub::BLOCK_SCAN_WARP_SCANS>;
+  };
+
+  using MaxPolicy = Policy350;
+};
+#endif // !TUNE_BASE
+
+#include <cub/device/device_run_length_encode.cuh>
```
miscco, line 38: Ditto include after code

benchmarks/bench/segmented_sort/small/keys.cu
```diff
@@ -0,0 +1,240 @@
...
+                                       TUNE_M_LOAD_ALGORITHM,
+                                       TUNE_M_LOAD_MODIFIER>>;
+  };
+
+  using MaxPolicy = Policy350;
+};
+#endif // !TUNE_BASE
+
+#include <fstream>
+#include <thrust/host_vector.h>
```
miscco, line 126: Will stop commenting

miscco (approved):

benchmarks/bench/reduce/by_key.cu
senior-zero: That matched some strange mental model of mine :smile: After thinking about this, you are right, includes should go before, thanks!

elstehle (approved):

PR #704, opened 2023-05-26
Initial port to new monorepo build system.

allisonvacanti:


CMakeLists.txt
```diff
@@ -1,26 +1,25 @@
...
+# 3.21 is the minimum for the developer build.
 cmake_minimum_required(VERSION 3.15)
 
-# Remove this when we use the new CUDA_ARCHITECTURES properties.
-if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.18)
-  cmake_policy(SET CMP0104 OLD)
-endif()
-
 # CXX is only needed for AppendOptionIfAvailable.
 project(CUB NONE)
```
robertmaynard, line 6:
```suggestion
project(CUB LANGUAGES NONE)
```

cmake/CubBuildCompilerTargets.cmake
```diff
@@ -94,7 +94,7 @@ function(cub_build_compiler_targets)
   foreach (cxx_option IN LISTS cxx_compile_options)
     target_compile_options(cub.compiler_interface INTERFACE
       $<$<COMPILE_LANGUAGE:CXX>:${cxx_option}>
-      $<$<COMPILE_LANG_AND_ID:CUDA,NVCXX>:${cxx_option}>
+      $<$<COMPILE_LANG_AND_ID:CUDA,NVCXX>:${cxx_option}> # TODO check that this works with the new nvc++ CMake recipe
```
robertmaynard, line 97: This compiler vendor ID doesn't exist

cmake/CubUtilities.cmake
```diff
@@ -1,13 +1,29 @@
...
+  if (NOT DEFINED cub_cuda_RDC)
+    message(AUTHOR_WARNING "RDC option required for cub_configure_cuda_target.")
+  endif()
+
+  if (cub_cuda_RDC)
     set_target_properties(${target_name} PROPERTIES
-      COMPILE_FLAGS "-gpu=rdc"
-    )
-  elseif ("Clang" STREQUAL "${CMAKE_CUDA_COMPILER_ID}")
+      CUDA_ARCHITECTURES "${CUB_CUDA_ARCHITECTURES_RDC}"
```
robertmaynard, line 22: set PIC to on as well.

examples/CMakeLists.txt
```diff
@@ -40,6 +32,7 @@ function(cub_add_example target_name_var example_name example_src cub_target)
   add_executable(${example_target} "${example_src}")
   target_link_libraries(${example_target} ${cub_target})
   cub_clone_target_properties(${example_target} ${cub_target})
+  cub_configure_cuda_target(${example_target} RDC ${CUB_FORCE_RDC})
```
robertmaynard, line 35: I personally don't like examples that can't be built with an installed version of a project. I would recommend dropping usage of `cub_clone_target_properties` and `cub_configure_cuda_target`

cmake/CubBuildCompilerTargets.cmake
robertmaynard: Plus lets use ` $<COMPILE_LANG_AND_ID:CUDA,...>` to be consistent with the rest of the file

examples/CMakeLists.txt
allisonvacanti: This is a good consideration, but outside of the scope of this PR and is an existing issue. We can open an issue.

cmake/CubBuildCompilerTargets.cmake
allisonvacanti: NVHPC is the new one

cmake/CubBuildTargetList.cmake
```diff
@@ -66,15 +66,7 @@ function(cub_set_target_properties target_name dialect prefix)
...
-    # CMake still emits errors about empty CUDA_ARCHITECTURES when CMP0104
-    # is set to OLD. This suppresses the errors for good.
-    if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.18)
-      set_target_properties(${target_name}
-        PROPERTIES
-          CUDA_ARCHITECTURES OFF
-      )
-    endif()
-
+    # TODO is this still needed?
```
allisonvacanti, line 69: Rob says no.

cmake/CubCudaConfig.cmake
```diff
@@ -1,87 +1,23 @@
...
-)
-
-option(CUB_ENABLE_EXAMPLES_WITH_RDC
-  "Build all CUB examples with RDC; examples which require RDC are not affected by this option."
-  ${option_init}
-)
+# Create a new arch list that only contains arches that support CDP:
+set(CUB_CUDA_ARCHITECTURES ${CMAKE_CUDA_ARCHITECTURES})
+set(CUB_CUDA_ARCHITECTURES_RDC ${CUB_CUDA_ARCHITECTURES})
+list(FILTER CUB_CUDA_ARCHITECTURES_RDC EXCLUDE REGEX "53|62|72|90")
```
allisonvacanti, line 10: This will break for `native|all|all-major` under certain conditions.

allisonvacanti: For now, we'll emit an error advising the user to explicitly specify arches or turn off `ENABLE_RDC` options.

We can also do introspection ourselves at some point to determine this if needed.

cmake/CubBuildCompilerTargets.cmake
allisonvacanti: Updated to use the NVHPC id and `COMPILER_LANG_AND_ID` genex consistently.

cmake/CubUtilities.cmake
allisonvacanti: Added `POSITION_INDEPENDENT_CODE ON` for all RDC targets.

test/CMakeLists.txt
```diff
@@ -144,6 +136,11 @@ function(cub_add_test target_name_var test_name test_src cub_target)
   set(config_meta_target ${config_prefix}.tests)
 
   if (is_catch2_test)
+    set(use_rdc_for_catch2_utils OFF)
+  if (CUB_ENABLE_RDC_TESTS OR CUB_FORCE_RDC)
+    set(use_rdc_for_catch2_utils ON)
+  endif()
```
senior-zero, line 142: Looks like else branch:
```suggestion
    if (CUB_ENABLE_RDC_TESTS OR CUB_FORCE_RDC)
      set(use_rdc_for_catch2_utils ON)
    endif()
```

Issue #695, opened 2023-05-22
Make decoupled look-back delay part of tuning


senior-zero:
Preliminary tuning of rle, select, partition, scan, *by_key algorithms shows that delays in decoupled look-back spin-loops depend on the data type and the algorithm. Before tuning these algorithms, we should make `CUB_DETAIL_L2_BACKOFF_NS` and `CUB_DETAIL_L2_WRITE_LATENCY_NS` part of algorithm-dependent tuning policy.

senior-zero:
@IlyaGrebnov thank you for your comment! 

I remember trying exponential back off, and it was always worse than static delay. I think it might be related to the effect you mentioned, so I'll try it again as part of tuning infrastructure to figure out it there's a different CTA size / items per thread configuration that allows exponential back off to outperform static delay. 

Regardless of the results, exponential back off is not applicable to `CUB_DETAIL_L2_WRITE_LATENCY_NS`, so it would have to be extracted into tuning policy.

PR #710, opened 2023-06-05
Introduce delay policy to decoupled look-back based algorithms

senior-zero:
This PR addresses the following issue https://github.com/NVIDIA/cub/issues/695. It also fixed unique by key policy misuse that led to 2x code generation. There should be no difference in generated assembly at this PR (except for unique by key). Batched memcpy is out of the scope for this PR. Actual tuning of the delay policy will be provided as a follow-up PR as well.


benchmarks/bench/scan/exclusive/base.cuh
```diff
@@ -25,67 +25,62 @@
...
+    NOMINAL_BLOCK_THREADS_4B,
+    NOMINAL_ITEMS_PER_THREAD_4B,
+    ComputeT,
+    LOAD_ALGORITHM,
+    LOAD_MODIFIER,
+    STORE_ALGORITHM,
+    SCAN_ALGORITHM,
+    cub::MemBoundScaling<NOMINAL_BLOCK_THREADS_4B, NOMINAL_ITEMS_PER_THREAD_4B, ComputeT>,
+    delay_constructor_t>;
+
```
miscco, line 66: Is there any place where we use this?

In all other policies we just add the delay_constructor_t

miscco (approved):

benchmarks/bench/scan/exclusive/base.cuh
senior-zero: There are no other places like this because other places do not have `MemBoundScaling` before delay constructor. This alias ensures that the threads / items that go into scaling are the same as the ones that are passed to scan policy.

elstehle (approved):

PR #738, opened 2023-09-07
64-bit indexing for DeviceSegmentedReduce

jecs:
- Added NumSegmentsT parameter to methods, so now num_segments can be of any integral type
- Added OffsetT type resolution in a manner similar to DeviceReduce: set to std::common_type_t of BeginOffsetIterT::value_type and EndOffsetIterT::value_type. (see detail/choose_offset.cuh)
- In a manner similar to DeviceReduce, uses static_cast<OffsetT> to convert num_segments before dispatching
- Makes use of std::common_type_t, std::void_t, and one inline constexpr variable (is_iterator_v).

[If needed, use of _t and _v features can be removed to make the code C++11 compliant. Doc texts need to be updated as well.]

### DIFF
--- a/cub/detail/choose_offset.cuh
+++ b/cub/detail/choose_offset.cuh
@@ -56,6 +56,36 @@ struct ChooseOffsetT
                                          unsigned long long>::type;
 };
 
+template <typename T,typename = void>
+struct is_iterator : std::false_type {};
+template <typename T>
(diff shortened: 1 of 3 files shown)


jecs:
Oh, I just noticed that you opened a new repository. Should I submit this pull request over there?

senior-zero:
@jecs yes, please open the PR against https://github.com/NVIDIA/cccl

jecs:
Done. Closing this request.

